PR #6, opened 2024-03-13
Update Readme

aditijannu:
- Updates README to be more informative about the build process
- Include steps for the enclave kernel build process
- Provide a script to setup the environment with dependencies and build the converter docker image


docker/amzn-linux-nbd/build-enclave-kernel.sh
```diff
@@ -0,0 +1,59 @@
...
+
+  # Copy the enclave kernel config file into the kernel repository
+  # The original file is available here:
+  # https://github.com/aws/aws-nitro-enclaves-cli/blob/main/blobs/x86_64/bzImage.config
+  # The config file available in this directory has been updated to
+  # support salmiac features.
+  cp ../bzImage.config .config
+
+  # Build the enclave kernel
+  make prepare && make modules_prepare && make modules && make && make bzImage
```
NirjharRoyiitk, line 36: there are almost 4 commands here and all failure in one of them will 
1- Make it difficult to figure out which one has failed
2- The user might copy the entire command and re run it which will make all the sub commands to re run. 
Can we separate them?

docker/amzn-linux-nbd/build-enclave-kernel.sh
```diff
@@ -0,0 +1,59 @@
...
+  if [ -d "linux" ]; then
+    rm -rf linux
+  fi
+}
+
+if [ $1 == "build" ]; then
+  clonelinux;
+  buildkernel;
+elif [ $1 == "clean" ]; then
+ cleankernel;
```
NirjharRoyiitk, line 55: How about we call the cleankernel function as a signal on error just like the way ``trap cleankernel EXIT``

https://bitbucket.org/fortanix/zircon-apps/src/3618f1d34bacfe4ebe7fac14557c107d17e6cb7b/build.sh#lines-192 --> This is a sample. THis will make sure on some sudden exit we don't leave unused contents.

docker/build-conv-container.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Move the converter binary into staging, the input argument to the script
+# determines whether a release or debug converter is used
+cp ../tools/container-converter/target/$1/container-converter ./staging
+mv ./staging/container-converter ./staging/server
+
+# Package the enclave and parent base images for the converter
+docker save -o ./staging/nitro-enclave-base.tar enclave-base
+docker save -o ./staging/nitro-parent-base.tar parent-base
+
```
NirjharRoyiitk, line 15: Minor:
Can we have directory names/paths stored in some global variables with some brief descriptions of the things the paths store.

README.md
```diff
@@ -20,51 +20,64 @@ Useful links
 Quick Start Guide
 --------------
 
-1. Install Rust:
```
NirjharRoyiitk, line 23: Can we have section on debugging or at least run some of the CI tests/regression tests or unit tests?

docker/amzn-linux-nbd/build-enclave-kernel.sh
MihirLuthra: Can we add `-j <get_number_of_cpus>` flag here as well so that make uses multiple jobs? Or has that already been set somewhere?

docker/amzn-linux-nbd/build-enclave-kernel.sh
```diff
@@ -0,0 +1,59 @@
...
+  # by running uname -a in a converted app. In this case, we
+  # use version 4.14.246
+  # For more details about supported nitro enclave kernel
+  # versions, refer to the aws-nitro-enclaves-cli github repository.
+
+  if [ -d "linux" ]; then
+    cd linux
+    git fetch
+  else
+    git clone https://github.com/amazonlinux/linux.git
```
MihirLuthra, line 19: missing `cd` after this

docker/amzn-linux-nbd/build-enclave-kernel.sh
aditijannu: Building the kernel is a one time setup step and is not expected to run as periodically. Having the trap clean up statement will cause the linux repository to be cleaned up and prevent a user from debugging any errors or failures seen during the kernel build. I would say that the clean up should be a manual step when someone has finished building the kernel and obtained the artifacts as needed.

README.md
```diff
@@ -20,51 +20,64 @@ Useful links
...
+   ./build-enclave-kernel.sh build
+    ```
+
+5. Build the converter image. To produce a debug build of the converter, ensure the release flag is removed from the step below.
     ```bash
       # Run from the root of the repository
       cd salmiac
-      ./build-converter.sh   
+      # To produce a debug build of the converter, ensure the release flag is removed from the step below.
+      ./build-converter.sh --release
```
MihirLuthra, line 61: Building this needs following packages:
```
sudo apt install \
    pkg-config \
    libclang-dev \
    cmake \
    libpcap-dev
```

docker/amzn-linux-nbd/build-enclave-kernel.sh
aditijannu: I think the kernel build steps should happen sequentially. Why do you want to add the `-j` option?

MihirLuthra: If there is something in the Makefile that relies on it being built sequentially, then we can't add `-j`. You may have better insight there. Generally, `Makefile`s are smart enough to parallelize non-dependent jobs only. But yea, that isn't always the case.
If it was parallelizable, the time taken to build would reduce a lot.

README.md
aditijannu: Are you sure we need those tools to build the converter? Rust should be enough to build the converter.

MihirLuthra: Code in build scripts of some dependencies needed these.

docker/amzn-linux-nbd/build-enclave-kernel.sh
arai-fortanix: The plain `make` and `make modules` should have `-j` added, since those are the steps that take the longest to run. It won't help as much for `make prepare`, `make modules_prepare` or `make bzImage`.

README.md
```diff
@@ -19,52 +19,66 @@ Useful links
 
 Quick Start Guide
 --------------
+This guide allows you to build salmiac from source and convert your docker application into a one that can run in a nitro enclave. 
 
-1. Install Rust:
-   Follow [this](https://www.rust-lang.org/tools/install) guide.
+1. Set up your Ubuntu based build system:
+    - Install Rust:
```
arai-fortanix, line 25: Do we need to specify what version of the rust toolchain is required?

arai-fortanix (approved):

README.md
aditijannu: The build-converter.sh script adds the toolchain targets it needs with the help of rustup.

docker/build-conv-container.sh
aditijannu: Created https://fortanix.atlassian.net/browse/SALM-566 to address this in the future

README.md
aditijannu: Created https://fortanix.atlassian.net/browse/SALM-565 to address debugging in the future

README.md
```diff
@@ -73,19 +95,19 @@ Quick Start Guide
...
    ```
 
-7. Make your application Nitro VM-capable by running container converter with the file from previous step
+7. Make your application Nitro VM-capable by running container converter with the file from previous step.
+   The converter by default pulls the input image and pushes the output image to remote repositories. These images are then cleaned up from the local docker cache. In our example, the output image push is disabled in the request json and to preserve the images in the docker cache, 'PRESERVE_IMAGES' environment variable is specified.
    ```bash
-      # Run from the root of the repository
-      cd tools/container-converter/target/debug
-      ./container-converter --request-file <path to file from step 4>
+      docker run --rm --name converter --user 0 --privileged -v /var/run/docker.sock:/var/run/docker.sock -e PRESERVE_IMAGES=input,result -v /tmp/req-files:/app converter --request-file /app/req.json
```
NirjharRoyiitk, line 101: Minor: Having the parameters in separate lines looks better and increases readability 

 docker run --rm 
    --name converter 
    --user 0 --privileged 
    -v /var/run/docker.sock:/var/run/docker.sock 
    -e PRESERVE_IMAGES=input,result 
    -v /tmp/req-files:/app converter 
    --request-file /app/req.json

NirjharRoyiitk (approved):

README.md
aditijannu: Since this is part of the README and inside the bash brackets, it will be available in a scrollable dialog box with a copy button. I think it is usable enough for now.

NirjharRoyiitk (approved):

nshyrei (approved):

PR #11, opened 2024-09-03
Check application configuration hash when requesting application configuration

nshyrei:
This PR returns back functionality to check if the application configuration struct from CCM is indeed the one that we expect to see when starting a user application inside Salmiac. The verification process involves comparing the `SHA256` hash value of the `application configuration` structure with the `SHA256` hash value passed by the user through `ENCLAVEOS_APPCONFIG_ID` or `APPCONFIG_ID` env var when starting Salmiac itself.

This hash check prevents  malicious application configurations from being used when running a user application. Failure to pass the hash check results in no user application being run and Salmiac exiting with error code `-1`

### DIFF
--- a/vsock-proxy/enclave/src/app_configuration.rs
+++ b/vsock-proxy/enclave/src/app_configuration.rs
@@ -8,7 +8,7 @@ use std::collections::BTreeMap;
 use std::fs;
 use std::path::{Component, Path, PathBuf};
 use std::sync::Arc;
-use std::convert::TryFrom;
+use std::convert::TryInto;
 
 use em_app::utils::models::{
     ApplicationConfigContents, ApplicationConfigExtra, ApplicationConfigSdkmsCredentials, RuntimeAppConfig,
@@ -57,7 +57,10 @@ where
 {
     info!("Requesting application configuration.");
 
-    let app_config_id_raw = <[u8; APPLICATION_CONFIGURATION_ID_LENGTH]>::try_from(app_config_id.as_bytes()).map_err(|err| format!("Cannot convert application config id into an array. Error: {:?}. Perhaps your id isn't {} characters long?", err, APPLICATION_CONFIGURATION_ID_LENGTH))?;
+    let app_config_id_raw: [u8; APPLICATION_CONFIGURATION_ID_LENGTH] = app_config_id
+        .as_bytes()
+        .try_into()
+        .map_err(|err| format!("Cannot convert application config id into an array. Error: {:?}. Perhaps your id isn't {} characters long?", err, APPLICATION_CONFIGURATION_ID_LENGTH))?;
 
     let app_config = api
         .runtime_config_api()
@@ -673,7 +676,7 @@ mod tests {
         let credentials = EmAppCredentials::mock();
         let api: Box<dyn RuntimeConfiguration> = Box::new(MockDataSet { json_data, hash: [0; APPLICATION_CONFIGURATION_ID_LENGTH] });
 
-        let result = api.get_checked_runtime_configuration(&backend_url, &credentials, &[0; APPLICATION_CONFIGURATION_ID_LENGTH]);
+        let result = api.get_runtime_configuration(&backend_url, &credentials, &[0; APPLICATION_CONFIGURATION_ID_LENGTH]);
         assert!(result.is_ok(), "{:?}", result);
 
         result.unwrap()
--- a/vsock-proxy/enclave/src/app_configuration.rs
+++ b/vsock-proxy/enclave/src/app_configuration.rs
@@ -446,8 +446,7 @@ mod tests {
     use sdkms::api_model::Blob;
     use shared::models::CCMBackendUrl;
 
-    use crate::app_configuration::{normalize_path_and_make_relative, setup_app_configs, setup_datasets, ApplicationFiles, DataSetFiles, EmAppCredentials, RuntimeConfiguration, SdkmsDataset, ApplicationConfiguration, check_application_config_id};
-    use em_app::compute_app_config_hash;
+    use crate::app_configuration::{normalize_path_and_make_relative, setup_app_configs, setup_datasets, ApplicationFiles, DataSetFiles, EmAppCredentials, RuntimeConfiguration, SdkmsDataset, ApplicationConfiguration};
 
     const TEST_FOLDER: &'static str = "/tmp/salm-unit-test";
 
@@ -580,6 +579,7 @@ mod tests {
 
     struct MockDataSet {
         pub json_data: &'static str,
+        pub hash: Blob
     }
 
     impl MockDataSet {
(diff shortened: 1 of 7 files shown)


vsock-proxy/enclave/src/app_configuration.rs
```diff
@@ -44,29 +43,25 @@ const CREDENTIALS_FILE: &str = "credentials.bin";
...
     fs_root: &Path,
-    app_config_id: Blob
+    app_config_id: &str
 ) -> Result<(), String>
 where
     T: ApplicationConfiguration,
 {
     info!("Requesting application configuration.");
 
+    let app_config_id_raw = <[u8; APPLICATION_CONFIGURATION_ID_LENGTH]>::try_from(app_config_id.as_bytes()).map_err(|err| format!("Cannot convert application config id into an array. Error: {:?}. Perhaps your id isn't {} characters long?", err, APPLICATION_CONFIGURATION_ID_LENGTH))?;
```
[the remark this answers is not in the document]
nshyrei: I understand your confusion and I initially had this field as a byte array, but then I have found another function in `em-app` that is used in Salmiac and it requires a `&str`. The function in question is: https://github.com/fortanix/rust-sgx/blob/d446f203a6929d85250d5f21fe0e4bcf2f91efb9/em-app/src/lib.rs#L110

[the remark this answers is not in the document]
nshyrei: The code change you propose doesn't work because `try_from` is an associated function.
```
error[E0599]: no method named `try_from` found for reference `&[u8]` in the current scope
   --> enclave/src/app_configuration.rs:61:10
    |
60  |       let app_config_id_raw: [u8; APPLICATION_CONFIGURATION_ID_LENGTH] = app_config_id.as_bytes()
    |  ________________________________________________________________________-
61  | |         .try_from()
    | |         -^^^^^^^^ this is an associated function, not a method
    | |_________|
    |
    |
    = note: found the following associated functions; to be used as methods, functions must have a `self` parameter
note: the candidate is defined in the trait `TryFrom`
   --> /home/nikitashyrei/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/convert/mod.rs:687:5
    |
687 |     fn try_from(value: T) -> Result<Self, Self::Error>;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: use associated function syntax instead
    |
60  |     let app_config_id_raw: [u8; APPLICATION_CONFIGURATION_ID_LENGTH] = [u8]::try_from()
    |                                                                        ~~~~~~~~~~~~~~~~
help: there is a method `from` with a similar name
    |
61  |         .from()
    |          ~~~~


```

vsock-proxy/enclave/src/app_configuration.rs
```diff
@@ -17,12 +18,10 @@ use mbedtls::alloc::List as MbedtlsList;
 use mbedtls::pk::Pk;
 use mbedtls::x509::Certificate;
 use sdkms::api_model::Blob;
-use shared::models::CCMBackendUrl;
+use shared::models::{CCMBackendUrl};
```
aditijannu, line 21: Run cargo fmt on files that were modified.

vsock-proxy/enclave/src/app_configuration.rs
```diff
@@ -44,29 +43,28 @@ const CREDENTIALS_FILE: &str = "credentials.bin";
...
+    let app_config_id_raw: [u8; APPLICATION_CONFIGURATION_ID_LENGTH] = app_config_id
+        .as_bytes()
+        .try_into()
+        .map_err(|err| format!("Cannot convert application config id into an array. Error: {:?}. Perhaps your id isn't {} characters long?", err, APPLICATION_CONFIGURATION_ID_LENGTH))?;
+
     let app_config = api
         .runtime_config_api()
-        .get_runtime_configuration(&ccm_backend_url, em_app_credentials)?;
-
-    let _app_config_id_check = check_application_config_id(&app_config.config, app_config_id);
```
aditijannu, line 63: I'm having a hard time figuring out where the config hash is verified if this call is removed?

aditijannu: Ah nevermind, I see the check is made in `em-client-rust` crate

vsock-proxy/enclave/src/app_configuration.rs
```diff
@@ -44,57 +44,58 @@ const CREDENTIALS_FILE: &str = "credentials.bin";
...
 
+const APPLICATION_CONFIGURATION_ID_LENGTH: usize = 32;
+
 pub(crate) fn setup_application_configuration<T>(
     em_app_credentials: &EmAppCredentials,
     ccm_backend_url: &CCMBackendUrl,
     api: T,
     fs_root: &Path,
-    app_config_id: Blob
+    app_config_id: &str,
```
raoulstrackx, line 54: This is odd. A `Blob` (and the rest of the code) makes this assume the raw sha256 values. A `&str` makes me think this sha256 value is encoded in some way (e.g., base64).

vsock-proxy/enclave/src/app_configuration.rs
```diff
@@ -21,8 +23,6 @@ use shared::models::CCMBackendUrl;
 
 use crate::certificate::CertificateResult;
 use crate::enclave::write_to_file;
-use em_app::compute_app_config_hash;
-use em_app::utils::models;
```
raoulstrackx, line 25: There are a lot of changes due to `cargo fmt`. Can you split it up in two PRs for clarity?

nshyrei: Aditi wanted me to run the fmt https://github.com/fortanix/salmiac/pull/11#discussion_r1796899509, your request kinda contradicts her request.

aditijannu: I'm okay to move cargo fmt changes to another PR or if it is a separate commit in this PR.

nshyrei: It is a separate commit

vsock-proxy/enclave/src/app_configuration.rs
```diff
@@ -44,12 +43,47 @@ const CREDENTIALS_FILE: &str = "credentials.bin";
...
 
+// https://en.wikipedia.org/wiki/SHA-2
+const SHA256_BYTE_LENGTH: usize = 32;
+
+const SHA256_CHAR_LENGTH: usize = SHA256_BYTE_LENGTH * 2;
+
+#[derive(Debug)]
+pub struct Sha256Hash([u8; SHA256_BYTE_LENGTH]);
+
+impl TryFrom<&str> for Sha256Hash {
```
aditijannu, line 54: em-client had the exact same code, is it possible to reuse the code since we are anyway using em-app here rather than duplicate this?

vsock-proxy/enclave/src/app_configuration.rs
```diff
@@ -340,6 +321,7 @@ pub(crate) trait RuntimeConfiguration {
         &self,
         ccm_backend_url: &CCMBackendUrl,
         credentials: &EmAppCredentials,
+        expected_hash: Sha256Hash,
```
raoulstrackx, line 324: There's no need to pass this with ownership. Can you pass this by reference?

vsock-proxy/enclave/src/app_configuration.rs
```diff
@@ -49,7 +48,7 @@ pub(crate) fn setup_application_configuration<T>(
     ccm_backend_url: &CCMBackendUrl,
     api: T,
     fs_root: &Path,
-    app_config_id: Blob
+    app_config_id: Sha256Hash
```
raoulstrackx, line 51: Can you turn this into a reference as well?

raoulstrackx (approved):

PR #12, opened 2024-09-19
Run build rust code CI only when PR is approved

nshyrei:
This change changes CI behavior from running every time a new change is pushed into the branch to running only when PR is approved. This will decrease the number of CI runs as well as introduce a security guarantee that nobody can potentially extract CI secrets through malicious CI run by updating the PR with bogus data.

### DIFF
--- a/.github/workflows/build-solution.yml
+++ b/.github/workflows/build-solution.yml
@@ -15,10 +15,9 @@ on:
         required: true
       PAT:
         required: true
-  push:
-    branches: [ "master" ]
-  pull_request:
-    branches: [ "master" ]
+  on:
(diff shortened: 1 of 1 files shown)


nshyrei:
I believe that `build-docker-images.yml` workflow shouldn't be changed as it uses secrets only when a dependent PR is merged.

raoulstrackx:
I'll accept and merge the PR immediately, but someone else take another look if this suffices.

raoulstrackx:
Would it be an option to split the workflow such that not every CI run pushes to the ECR, and turn that step into some manual action?

PR #13, opened 2024-09-24
Change ECR buffer path

nshyrei:
Changes the ECR buffer used to transfer application tests container into a Nitro-capable EC2 instance to a new path.

### DIFF
--- a/build-app-tests-container.sh
+++ b/build-app-tests-container.sh
@@ -10,7 +10,7 @@ FLAVOR=debug
 if [ -z "$SKIP_RUNNING_TESTS" ]; then
   make tests-container FLAVOR=$FLAVOR
   TESTS_CONTAINER_TAG=$(cat build/nitro-$FLAVOR/tests-container-tag)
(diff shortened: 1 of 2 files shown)


build-app-tests-container.sh
```diff
@@ -10,7 +10,7 @@ FLAVOR=debug
 if [ -z "$SKIP_RUNNING_TESTS" ]; then
   make tests-container FLAVOR=$FLAVOR
   TESTS_CONTAINER_TAG=$(cat build/nitro-$FLAVOR/tests-container-tag)
-  TESTS_CONTAINER_ECR="513076507034.dkr.ecr.us-west-1.amazonaws.com/development-images/$TESTS_CONTAINER_TAG"
+  TESTS_CONTAINER_ECR="513076507034.dkr.ecr.us-west-1.amazonaws.com/salmiac-github-ci/$TESTS_CONTAINER_TAG"
```
aditijannu, line 13: Do we use this script in salmiac-release job or a different script?

Can we ensure that github artifacts are not used by the salmiac-release job?

nshyrei: I don't think we use it anywhere else

PR #16, opened 2024-10-02
Turn off application tests CI

nshyrei:
This PR turns off application tests CI until we move test images into a public repository. Additionally the PR contains the following small additions:
1) Reverted back policy to run CI only on successful review as it blocks the iterative process to test changes for CI itself. Now with application tests turned off there is no danger in anyone running a small package of `project build + unit tests`
2) `docker-credential-ecr-login` is now downloaded from a public place instead of a private one

### DIFF
--- a/.github/workflows/build-solution.yml
+++ b/.github/workflows/build-solution.yml
@@ -15,9 +15,10 @@ on:
         required: true
       PAT:
         required: true
-  on:
-    pull_request_review:
-      types: [ submitted, edited ]
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
 
 env:
   CARGO_TERM_COLOR: always
@@ -26,7 +27,7 @@ jobs:
   build:
 
     runs-on: ubuntu-20.04
-    if: github.event_name == 'pull_request' && github.event.review.state == 'approved'
+
     steps:
       - name: Pull main repository
         uses: actions/checkout@v3
@@ -67,16 +68,8 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libpcap-dev amazon-ecr-credential-helper
 
-      - name: Install docker-credential-ecr-login
-        run: |
-          # Download the latest release of docker-credential-ecr-login
-          aws s3 cp s3://fortanix-internal-artifact-repository/other-packages/docker-credential-ecr-login .
-
-          # Make it executable
-          chmod +x docker-credential-ecr-login
-
-          # Move it to a directory in the PATH
-          sudo mv docker-credential-ecr-login /usr/local/bin/
+          # Setup docker-credential-ecr-login
+          sudo mv /usr/bin/docker-credential-ecr-login /usr/local/bin/
 
       - name: Build Rust code
         run: |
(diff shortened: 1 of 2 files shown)


aditijannu:
Why is the branch named SALM-213? Can we rename the branch to the relevant ticket?

.github/workflows/build-solution.yml
```diff
@@ -67,16 +68,8 @@ jobs:
...
-        run: |
-          # Download the latest release of docker-credential-ecr-login
-          aws s3 cp s3://fortanix-internal-artifact-repository/other-packages/docker-credential-ecr-login .
-
-          # Make it executable
-          chmod +x docker-credential-ecr-login
-
-          # Move it to a directory in the PATH
-          sudo mv docker-credential-ecr-login /usr/local/bin/
+          # Setup docker-credential-ecr-login
```
aditijannu, line 71: Can we add a more relevant comment here? Why are we moving the binary to /usr/local/bin from /usr/bin? `/usr/bin` is already part of the $PATH.

.github/workflows/build-solution.yml
```diff
@@ -89,13 +82,13 @@ jobs:
         run: |
           ./unit-test-solution.sh
 
-      - name: Build application tests container
-        run: |
-          ./build-app-tests-container.sh
+#      - name: Build application tests container
```
aditijannu, line 85: Add the todo comment to enable this step along with the jira ticket number that involves the todo task to be done.

.github/workflows/build-solution.yml
nshyrei: It's because of this rule https://github.com/fortanix/salmiac/blob/7f320a37b9f84e5a288c236cff707dd8ffdc43c3/make/salmiac-tests-container.make#L130.

nshyrei:
> Why is the branch named SALM-213? Can we rename the branch to the relevant ticket?

This is the correct ticket number - "Turn off Application tests in Salmiac"

aditijannu:
> > Why is the branch named SALM-213? Can we rename the branch to the relevant ticket?
> 
> This is the correct ticket number - "Turn off Application tests in Salmiac"

It's RTE-213 and not SALM-213

.github/workflows/build-solution.yml
```diff
@@ -15,9 +15,10 @@ on:
         required: true
       PAT:
         required: true
-  on:
-    pull_request_review:
-      types: [ submitted, edited ]
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
```
raoulstrackx, line 21: Nitpick: Personally I'm in strong favor of a one-goal-one-PR rule. It makes it much more clear why these changes are there. I agree to what you're trying to do, but doesn't that mean we can also disable the other keys here?

nshyrei: I don't fully understand what keys are you talking about here, do you want to remove the `push/pull` rules? I believe in that case CI won't start.

.github/workflows/build-solution.yml
```diff
@@ -89,13 +84,14 @@ jobs:
...
           ./unit-test-solution.sh
 
-      - name: Build application tests container
-        run: |
-          ./build-app-tests-container.sh
-
-      - name: Run application tests container
-        env:
-          FORTANIX_API_KEY: ${{ secrets.FORTANIX_API_KEY }}
-          OVERLAYFS_UNIT_TEST_API_KEY: ${{ secrets.OVERLAYFS_UNIT_TEST_API_KEY }}
```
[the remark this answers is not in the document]
nshyrei: They are still used in https://github.com/fortanix/salmiac/blob/af3ad673dc6e5b8c282d6a6fa0f9ee997829f487/.github/workflows/build-solution.yml#L80

PR #19, opened 2024-10-22
SALM-572 Update converter API to include CCM/DSM configs

aditijannu:
- Converter API is updated to include two additional converter options which specify the endpoint for nitro enclaves to connect to while requesting app configs and encryption keys
   - ccm_configuration
   - dsm_configuration
- Also ensure skip_server_verify and fs_api_key environment variables are only used in enclaves converted with the debug converter option


nshyrei (changes requested):
For some files there is a sign of a missing newline, can you fix that please?
image

aditijannu:
> For some files there is a sign of a missing newline, can you fix that please? image

Cargo fmt changes my PR quiet a bit, I'd like to do a cargo fmt at the end once review is complete. I also created RTE-267 to deal with this issue.

tools/container-converter/src/image_builder/enclave.rs
```diff
@@ -103,7 +107,9 @@ impl EnclaveSettings {
             user_name: input_image.details.config.user.clone(),
             env_vars: vec![rust_log_env_var("enclave")],
             is_debug: converter_options.debug.unwrap_or(false),
-            enable_overlay_filesystem_persistence: converter_options.enable_overlay_filesystem_persistence.unwrap_or(false)
+            enable_overlay_filesystem_persistence: converter_options.enable_overlay_filesystem_persistence.unwrap_or(false),
+            ccm_backend_url: CcmBackendUrl::new(converter_options.ccm_configuration.clone().unwrap_or_default().ccm_url.as_str()).unwrap_or_default(),
+            dsm_configuration: converter_options.dsm_configuration.clone().unwrap_or_default()
```
raoulstrackx, line 112: Nitpick: in Rust you can have a `,` at the end here. Then next time you add a field, no line needs to be removed just because it needs a `,`

vsock-proxy/enclave/src/dsm_key_config.rs
```diff
@@ -25,34 +25,38 @@ use sdkms::SdkmsClient;
...
 }
 
-fn dsm_create_client(env_vars: &[(String, String)], auth_cert: Option<&mut CertificateResult>) -> Result<SdkmsClient, String> {
-    info!("Looking for env variables needed to create DSM client.");
-    let api_key = find_env_or_err("FS_API_KEY", env_vars).unwrap_or("".to_string());
+/// Information needed to connect to DSM as a client
+pub struct ClientConnectionInfo<'a> {
+    pub fs_api_key: Option<String>,
+    pub auth_cert: Option<&'a mut CertificateResult>,
+    pub dsm_url: String,
```
raoulstrackx, line 43: Do these fields need to be `pub`? Having them as `pub` means they can be changed anywhere, at any point, even outside of this crate. That's not "wrong", but makes it more difficult to reason about them. Having only private fields, and a public constructor, makes it obvious that a value of this struct is created once, and then won't be changed.

vsock-proxy/enclave/src/file_system.rs
```diff
@@ -274,7 +270,7 @@ async fn get_key_file(
 /// Generate the luks2 token object and write the same to
 /// the json file which will be used to add a luks2 header
 /// to the RW blockfile
-fn create_luks2_token_input(token_path: &str, env_vars: &[(String, String)], enc_resp: EncryptResponse) -> Result<(), String> {
+fn create_luks2_token_input(token_path: &str, dsm_url: String, enc_resp: EncryptResponse) -> Result<(), String> {
```
raoulstrackx, line 273: Can you pass this by reference?

api-model/src/converter.rs
```diff
@@ -274,3 +280,33 @@ impl CertificateConfig {
...
+            ccm_url: "ccm.fortanix.com:443".to_string(),
+        }
+    }
+}
+
+/// Describes the information required to access DSM
+#[derive(Clone, Eq, PartialEq, Debug)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DsmConfiguration {
+    pub dsm_url: String,
```
raoulstrackx, line 303: I was expecting a `Url` here, not a `String`. I guess it's too late now as it passed API review?

raoulstrackx, line 303: Can you make this private?

aditijannu: Yes, but I do check at conversion time that it can be parsed as a Url else I throw an error.

aditijannu: Also we pass this parameter as a string to sdkms client so I think it's okay to leave it as a string.

aditijannu: I can't. It is used in the converter crate to check if it's a valid url.

vsock-proxy/enclave/src/dsm_key_config.rs
aditijannu: I can update it to pub(crate), they need to be accessible in the vsock-proxy/enclave crate only.

raoulstrackx (approved):

PR #20, opened 2024-10-22
use [patch] sections to specify serde_json dependency

tvsfx:
The repo currently uses explicit git dependencies for `serde_json` ([for example here](https://github.com/fortanix/salmiac/blob/4d3e1af07ca1a85bb3f633d8ae9b9a43cd8a71b5/api-model/Cargo.toml#L13))

This has to two disadvantages:
* It has the undesired side-effect of making these dependencies non-overridable in crates that depend on (crates in) this repo. This is similar to a `=`-version constraint in a `Cargo.toml` and can cause version collisions if dependents use a different version of `serde_json` (no interaction between both versions is then possible).
* It does not patch `serde_json` in transitive dependencies of this crate, and hence makes it so this crate uses multiple versions of `serde_json` (see the diffs in the `Cargo.lock` files for this PR)

The one small advantage is that it puts no burden on the user of this crate to patch `serde_json` _for this crate specifically_, but this is easily outweighed by the disadvantages mentioned above.

Better is to use a `[patch]` section, so that dependencies are patched transitively (if parts of this repo are used and built stand-alone) but can still be re-patched/overridden by its clients. This PR implements this change.


tvsfx:
CI fails because of https://github.com/serde-rs/json/issues/444, which stipulates usage of `serde_json > 1.0.25`. Will try upating patches to the `1.0.51` branch until my patch for `1.0.128` lands and we can port to that.

nshyrei:
@tvsfx Please resolve merge conflicts

nshyrei (approved):

PR #24, opened 2025-01-10
Adapt legacy code to new version of application tests infrastructure

nshyrei:
This PR alters the legacy behavior of `container-converter` to fit with the new version of `app-test-infra` (https://github.com/fortanix/app-test-infra) when running application tests. In summary the following was done:

1) Container converter now prints `json serialized` conversion response into the stdout. Previously the response was serialized as a simple string and it would break the code in: https://github.com/fortanix/app-test-infra/blob/5abda4112fb64559eb4e0064cf78dc7e4a29cb0a/python/test_app.py#L1005C1-L1008C1
2) Container converter now prints a special key phrase when printing the response to stdout. This phrase is used to find the response by app-test infra code in: https://github.com/fortanix/app-test-infra/blob/5abda4112fb64559eb4e0064cf78dc7e4a29cb0a/python/test_app.py#L1383
3) The unfinished legacy debug image logic is completely removed from container converter. Previously it was part of the idea for converter to create a special image with debug tooling and a distinct suffix when `debug` flag is set to true. This was never realized into a full feature and now breaks the application tests infrastructure code because images with debug prefixes are not supported there. As this feature was never properly implemented and was never needed that much I believe it's best to remove it entirely.
4) Changed path of `DOCKER-AWS-HELPER` to be consistent with tool installation path on the executor machine.

### DIFF
--- a/tools/container-converter/src/lib.rs
+++ b/tools/container-converter/src/lib.rs
@@ -73,7 +73,7 @@ const PARENT_IMAGE: &str = "parent-base";
 
 const ENCLAVE_IMAGE: &str = "enclave-base";
 
-const ENCLAVE_IMAGE_DEBUG: &str = "enclave-base-debug";
+const ENCLAVE_IMAGE_DEBUG: &str = "enclave-base";
 
 const DEFAULT_RSA_SIZE: u32 = 3072;
 const RSA_KEY_SIZES: [u32; 3] = [ 2048, DEFAULT_RSA_SIZE, 4096 ];
@@ -203,11 +203,12 @@ async fn run0(
 
         let enclave_base_image = get_enclave_base_image(&enclave_base_image_str).await?;
 
-        let user_program_config = if conversion_request.is_debug() {
+        let user_program_config = create_user_program_config(&conversion_request.request.converter_options, &input_image.image)?;
+        /*let user_program_config = if conversion_request.is_debug() {
             enclave_base_image.create_user_program_config()?
         } else {
             create_user_program_config(&conversion_request.request.converter_options, &input_image.image)?
-        };
+        };*/
 
         debug!("User program config is: {:?}", user_program_config);
 
--- a/tools/container-converter/src/lib.rs
+++ b/tools/container-converter/src/lib.rs
@@ -73,8 +73,6 @@ const PARENT_IMAGE: &str = "parent-base";
 
 const ENCLAVE_IMAGE: &str = "enclave-base";
 
-const ENCLAVE_IMAGE_DEBUG: &str = "enclave-base";
-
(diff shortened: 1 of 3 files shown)


tools/container-converter/src/main.rs
```diff
@@ -30,7 +30,10 @@ async fn main() -> Result<(), String> {
 
     match container_converter::run(request).await {
         Ok(response) => {
-            println!("{:?}", response);
+            let response_serialized = serde_json::to_string(&response)
+                .map_err(|err| format!("Failed serializing conversion request. {:?}", err))?;
+
+            println!("Successful nitro conversion: {:?}", response_serialized);
```
nshyrei, line 36: The string corresponds to a search prefix defined here: https://github.com/fortanix/app-test-infra/blob/5abda4112fb64559eb4e0064cf78dc7e4a29cb0a/python/test_app.py#L1383

aditijannu:
Which jira ticket requires these changes? 
app-test-infra code is shared between zircon and salmiac. These changes should ideally be tested in both projects. Otherwise it is likely going to break one or both projects.

Instead of removing the debug image logic completely, shouldn't we be fixing it in another ticket?

nshyrei:
> Which jira ticket requires these changes? app-test-infra code is shared between zircon and salmiac. These changes should ideally be tested in both projects. Otherwise it is likely going to break one or both projects.
> 
> Instead of removing the debug image logic completely, shouldn't we be fixing it in another ticket?

@aditijannu The changes are related to a jira story where I move application tests CI to jenkins (RTE-215). I don't understand how these changes will break ZIrcon if I only touch Salmiac-related code here. About the debug image - it also prevents me from running the app-tests currently as the suffix addition logic prevents the image to be found by app-test code. Also nobody has ever used debug images as this feature wasn't really complete.

tools/container-converter/src/main.rs
aditijannu: Oh I'm surprised it was working without the prefix strings matching.

tools/container-converter/src/lib.rs
```diff
@@ -194,20 +192,12 @@ async fn run0(
 
     info!("Building enclave image!");
     let nitro_image_result = {
-        let enclave_base_image_str = if conversion_request.is_debug() {
```
aditijannu, line 197: Please create a follow up jira ticket to handle debug images.

aditijannu (approved):

tools/container-converter/src/main.rs
nshyrei: It wasn't, the old version of app-test-infra didn't have the code mentioned.

PR #25, opened 2025-01-10
Filter "_" env var 

nshyrei:
Filters the `"_"` env variable list that is sent to the `enclave` by the `parent` before starting the client application. The `"_"` variable holds the last argument of the previous command according to: https://unix.stackexchange.com/questions/436615/when-is-an-environment-variable-of-a-bash-shell. This information is not particularly necessary for the enclave and potentially can be a security risk, thus it's a good idea to filter it out.

### DIFF
--- a/vsock-proxy/parent/src/parent.rs
+++ b/vsock-proxy/parent/src/parent.rs
@@ -824,4 +824,15 @@ mod tests {
         // Ensure the key was not removed
         assert!(runtime_env_vars.get("RUNTIME_VAR").is_some());
     }
+
+    #[test]
+    fn test_runtime_key_removed_when_values_are_the_same() {
+        let mut runtime_env_vars = HashMap::new();
+        runtime_env_vars.insert("MY_VAR".to_string(), "value".to_string());
+
+        filter_parent_env_from_runtime_envs(&mut runtime_env_vars, ("MY_VAR", "value"));
+
+        // Ensure the key was removed
+        assert!(runtime_env_vars.get("MY_VAR").is_none());
+    }
 }
--- a/vsock-proxy/parent/src/parent.rs
+++ b/vsock-proxy/parent/src/parent.rs
@@ -36,6 +36,7 @@ use crate::network::{
 };
 use crate::packet_capture::start_pcap_loops;
 use crate::ParentConsoleArguments;
+use std::collections::HashMap;
 
 const INSTALLATION_DIR: &str = "/opt/fortanix/enclave-os";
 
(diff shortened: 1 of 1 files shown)


vsock-proxy/parent/src/parent.rs
```diff
@@ -186,30 +187,51 @@ async fn send_enclave_exit(enclave_port: &mut AsyncVsockStream) -> Result<(), St
...
-    // Filter the environment variables sent from the parent. The ORIG_ENV_LIST_PATH file contains
-    // the list of variables which were set in the parent at conversion time. Rules for filtering:
-    //  - Always pass HOSTNAME
-    //  - Keep all environment variables which are new (i.e. not present in the parent at conversion
-    //    time)
-    //  - Keep environment variables which were present in the parent container but their values
-    //    have now been updated. The exception to this rule is for the PATH variable.
+    // "_" is a special var setup by bash that points to the currently executed binary
+    // to not leak the binary path into the enclave we remove this var explicitly.
+    runtime_vars.remove("_");
```
raoulstrackx, line 195: The bash man page states:
```
       _      At shell startup, set to the absolute pathname used to invoke the shell or shell script being executed as passed in the environment or argument list.
```
This seems to imply that there always should be a `_` environment variable. Is there a test in our CI to verify that this still works correctly?

jethrogb:
`_` should be set to what it would be set to if you were running your container without salmiac.

nshyrei:
> `_` should be set to what it would be set to if you were running your container without salmiac.

@jethrogb @raoulstrackx Does it mean that we shouldn't filter it ?

vsock-proxy/parent/src/parent.rs
```diff
@@ -186,30 +187,51 @@ async fn send_enclave_exit(enclave_port: &mut AsyncVsockStream) -> Result<(), St
...
     for line in reader.lines() {
         let env_line = line.map_err(|e| format!("Unable to read line from file {:?} : {:?}", orig_env_path, e))?;
         // Ill formed env variables will be ignored
-        let env_key_val = env_line.split_once("=").unwrap_or(("", ""));
-        if env_key_val.0 != "HOSTNAME" {
-            info!("Testing if {} does not exist or has been updated.", env_line);
-            runtime_vars.retain(|o| o.0 != env_key_val.0 || (o.0 != "PATH" && (o.0 == env_key_val.0 && o.1 != env_key_val.1)));
+        let conversion_time_env_var = env_line.split_once("=").unwrap_or(("", ""));
+
+        filter_runtime_env_vars(&mut runtime_vars, conversion_time_env_var);
```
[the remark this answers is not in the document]
nshyrei: `retain` was used before when the type of `runtime_vars` was a `Vec` and it resulted in linear search for every `conversion_time_env` var. Even though the number of `runtime_vars` shouldn't be big I decided to make the filter more efficient by using a `HashMap` with explicit `contains/remove` functions that use constant number of operations. `retain` for `HashMap` will perform same linear search as for `Vec` and will be less efficient.

vsock-proxy/parent/src/parent.rs
```diff
@@ -751,4 +774,54 @@ mod tests {
...
+        runtime_env_vars.insert("PATH".to_string(), "old_value".to_string());
+
+        filter_runtime_env_vars(&mut runtime_env_vars, ("PATH", "new_value"));
+
+        // Ensure the key was removed
+        assert!(runtime_env_vars.get("PATH").is_none());
+    }
+
+    #[test]
+    fn test_key_not_removed_when_values_dont_match() {
```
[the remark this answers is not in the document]
nshyrei: I have rephrased test names and test values to properly reflect the test intention, I think it should give a proper description what test cases the tests are testing now.

vsock-proxy/parent/src/parent.rs
```diff
@@ -186,30 +187,51 @@ async fn send_enclave_exit(enclave_port: &mut AsyncVsockStream) -> Result<(), St
...
+    let (conv_time_env_key, conv_time_env_val) = parent_env;
+
+    if conv_time_env_key != "HOSTNAME" {
+        info!("Testing if {:?} does not exist or has been updated.", parent_env);
+
+        match runtime_env_vars.get(conv_time_env_key) {
+            Some(value) if value == conv_time_env_val || conv_time_env_key == "PATH" => {
+                runtime_env_vars.remove(conv_time_env_key);
+            }
+            _ => {
```
aditijannu, line 230: Which use case does this block cover? Can we have a unit test for this?

nshyrei: Can you be more specific? GIthub doesn't mark the block precisely so I don't understand what block you are curious about, also the whole logic is moved, not redesigned from the previous solution with a `Vec`. If you are asking about the test case with a `PATH` it's: `test_runtime_path_key_is_removed` and for the same values it's tests case: `test_runtime_key_not_removed_when_values_dont_match`.

vsock-proxy/parent/src/parent.rs
```diff
@@ -751,4 +774,54 @@ mod tests {
...
+        assert_eq!(
+            runtime_env_vars.get("MY_VAR"),
+            Some(&"new_value".to_string())
+        );
+    }
+
+    #[test]
+    fn test_runtime_key_not_removed_when_not_present() {
+        let mut runtime_env_vars = HashMap::new();
+        runtime_env_vars.insert("RUNTIME_VAR".to_string(), "runtime_value".to_string());
```
aditijannu, line 820: MY_VAR should also be inserted with "value" as the value. We should check that it is removed on line 825.

vsock-proxy/parent/src/parent.rs
aditijannu: I was referring to the catch all block. I see the `test_runtime_key_not_removed_when_values_dont_match` unit test now.

aditijannu (dismissed):

aditijannu (approved):

PR #26, opened 2025-01-14
[rte-301] Code cleanup

raoulstrackx:


Taowyoo (approved):

unit-test-solution.sh
```diff
@@ -9,16 +9,16 @@ if [ -z "$SKIP_RUNNING_TESTS" ]; then
       exit 1
   fi
   unit_test_dirs="vsock-proxy/enclave
-  vsock-proxy/parent
-  tools/container-converter"
+      vsock-proxy/parent
+      tools/container-converter"
   for unit_test_dir in $unit_test_dirs
```
Taowyoo, line 14: it's better to use standard way to define string array.
to replace line 11-14:
```
unit_test_dirs=(
    "vsock-proxy/enclave"
    "vsock-proxy/parent"
    "tools/container-converter"
)
for unit_test_dir in "${unit_test_dirs[@]}"
```

Taowyoo (approved):

PR #27, opened 2025-01-17
[DEVOPS-5457] Upgrade dependencies version

Taowyoo:
Upgrade `sys-mount` and `tar` version to avoid dependency on old version of bindgen.
This helps fixing the bindgen vs clang incompatibility issue. See details in internal Jira ticket DEVOPS-5457.


aditijannu:
The converter does use loop mounts at conversion time to create and populate block devices, so I'm not sure if it's okay to remove the loopdev feature. 

I would like to do a CI run or QA validation on this branch before we merge it.

Taowyoo:
Thanks for telling me this.
Then I will change solution to "upgrade sys-mount version" instead.

nshyrei:
@Taowyoo Please update the title of your PR according to the new changes.

dlegalov-fortanix (approved):
Looks like a reasonable update, as long as it is tested to work.

gausk (commented):
LGTM

PR #28, opened 2025-01-27
Build fix

nshyrei:
Fixes build failure on the older versions of the compiler by downgrading the `home` crate and pinning the toolchain version.

### DIFF
--- a/tools/container-converter/Cargo.toml
+++ b/tools/container-converter/Cargo.toml
@@ -25,7 +25,7 @@ serde = { version = "1.0.127", features = ["derive"] }
 serde_json =  { git = "https://github.com/fortanix/serde-json.git", branch = "base64_bytes" }
 shiplift = { git = "https://github.com/fortanix/shiplift.git" }
 sys-mount = "3"
-tar = "0.4.39"
+tar = { git = "https://github.com/alexcrichton/tar-rs" }
 tempfile = "3.2.0"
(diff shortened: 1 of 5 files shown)


vsock-proxy/Cargo.lock
```diff
@@ -894,11 +894,11 @@ checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
 
 [[package]]
 name = "home"
-version = "0.5.11"
+version = "0.5.9"
```
nshyrei, line 897: This is the fix

tools/container-converter/Cargo.toml
```diff
@@ -25,7 +25,7 @@ serde = { version = "1.0.127", features = ["derive"] }
 serde_json =  { git = "https://github.com/fortanix/serde-json.git", branch = "base64_bytes" }
 shiplift = { git = "https://github.com/fortanix/shiplift.git" }
 sys-mount = "3"
-tar = "0.4.39"
+tar = { git = "https://github.com/alexcrichton/tar-rs" }
```
raoulstrackx, line 28: Why is this needed? Patching a crate from crates.io should be by using a patch section.

nshyrei: Oh yeah, patch is correct here, I forgot about the patch. I need to revert this because it crashes `tar` reading in container converter.

PR #34, opened 2025-03-31
[RTE-421] Update github workflows

aditijannu:
* Removes running app tests in an AWS EC2 machine. This means that the CI only covers building the project and running unit tests now. App tests will be run in jenkins jobs.

* Update the rust toolchain version used to build salmiac in github to the same version that is used in jenkins to avoid CI failures.


.github/workflows/build-solution.yml
```diff
@@ -3,18 +3,10 @@ name: Build Solution
...
-      AWS_ACCESS_KEY_ID:
-        required: true
-      AWS_SECRET_ACCESS_KEY:
-        required: true
-      EC2_PRIVATE_SSH_KEY:
-        required: true
       FORTANIX_API_KEY:
         required: true
       OVERLAYFS_UNIT_TEST_API_KEY:
         required: true
```
raoulstrackx, line 9: Are these two secrets still required?

raoulstrackx (dismissed):

.github/workflows/build-solution.yml
```diff
@@ -34,39 +26,16 @@ jobs:
...
-          toolchain: nightly
-
-      - name: Setup SSH
-        run: |
-          mkdir -p ~/.ssh
-          echo "${{ secrets.EC2_PRIVATE_SSH_KEY }}" > ~/.ssh/id_rsa
-          chmod 600 ~/.ssh/id_rsa
-          ssh-keyscan github.com >> ~/.ssh/known_hosts
+          targets: x86_64-unknown-linux-musl, x86_64-unknown-linux-gnu
+          toolchain: nightly-2024-11-21
```
tvsfx, line 33: Is installing the same toolchain that we use in Roche (minus the `dev` fixes) what fixed the CI issue? What was the problem in the end?

aditijannu: Yes. By specifying 'nightly` as the toolchain parameter, we were pulling the latest nightly compiler in every CI run and using it to build salmiac. The CI started breaking after this compiler version `rustc 1.87.0-nightly (3ea711f17 2025-03-09)` - I can't tell which feature exactly in this version of the compiler broke the CI.

To maintain similar builds between github CI and jenkins CI, I think it is better to pin the nightly version to the same version used in jenkins/roche.

.github/workflows/build-solution.yml
aditijannu: Yes they are used by the unit tests

.github/workflows/build-solution.yml
aditijannu: Refer to RTE-429 for the error observed with the `rustc 1.87.0-nightly (3ea711f17 2025-03-09)` build.

raoulstrackx (approved):

PR #37, opened 2025-05-06
Update workflows to run on ubuntu 24

aditijannu:


Taowyoo (approved):

aditijannu:
> @aditijannu approved. Do we have plan to upgrade toolhain so we can upgrade bindgen? Also I suggest to add **dependabot** to this repo to help maintain dependency upgrade automatically like we did in `rust-sgx`, see https://github.com/fortanix/rust-sgx/blob/master/.github/dependabot.yml

@Taowyoo Yes we have a plan to update the toolchain soon. Regarding the dependabot workflow, that's a good idea, created RTE-511 to track it.

PR #38, opened 2025-05-13
[RTE-431] Allow enclave to obtain certificate expiry from pem string

aditijannu:


vsock-proxy/enclave/src/certificate.rs
```diff
@@ -289,4 +299,23 @@ mod tests {
...
+        let mut test_resource = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        test_resource.push("resources/test");
+
+        let mut certpath = PathBuf::from(test_resource.clone());
+        certpath.push("valid_cert.pem");
+        let mut cert_contents: Vec<u8> = Vec::new();
+        let _cert_size = File::open(certpath)
+            .unwrap()
+            .read_to_end(&mut cert_contents)
+            .unwrap();
```
raoulstrackx, line 314: Nitpick: you could use `std::fs::read_to_string` here

raoulstrackx (approved):

PR #43, opened 2025-06-23
[RTE-483] Upgrade nitro-cli to v1.4.2 

aditijannu:
- Update nitro-cli to v1.4.2 in parent-base image
- Update parent-base image tag and it's references
- Update the custom enclave kernel build used by nitro-cli in the converter


raoulstrackx (dismissed):

docker/amzn-linux-nbd/build-enclave-kernel.sh
```diff
@@ -33,6 +33,11 @@ buildkernel() {
   # support salmiac features.
   cp ../bzImage.config .config
 
+  # For building kernel v4.14.256, the following edits need to be made
+  # append -Wno-use-after-free to the this line "CFLAGS += -ggdb3 -Wall -Wextra -std=gnu99 -fPIC"
```
raoulstrackx, line 37: Why is this needed?

raoulstrackx (dismissed):

docker/amzn-linux-nbd/build-enclave-kernel.sh
raoulstrackx: Oh this is about silencing use-after-free warnings that are treated as an error

AnjanHK (commented):
verified nitro-converted image execution it is successful

PR #45, opened 2025-07-02
[RTE-500] Prevent indefinite wait for app cert

raoulstrackx:
Bugfixes:
 - Requesting a new certificate from a background task is a sync call. That shouldn't happen directly from an async task.
 - The wait time for a new certificate should be limited. If it hasn't responded in time, it should time out and eventually try again
 - Upon an error the enclave should be notified. There was an issue where the result was ignored correctly, but the write operation on the vsock wasn't awaited.
 - Corrections in comments and error strings

### DIFF
--- a/vsock-proxy/parent/lib/lib.rs
+++ b/vsock-proxy/parent/lib/lib.rs
@@ -78,7 +78,7 @@ pub async fn handle_csr_message<Socket: AsyncWrite + AsyncRead + Unpin + Send, C
             // Failures may be silently dropped (i.e., when the enclave renews certificate in a
             // background task periodically. Ensure it can retry after some time and doesn't keep
             // waiting.
-            let _ = vsock.write_lv_bytes(&[]).await;
+            vsock.write_lv_bytes(&[]).await?;
             Err(e)
         },
     }


vsock-proxy/parent/lib/lib.rs
```diff
@@ -44,41 +48,65 @@ fn node_agent_address() -> Option<String> {
...
+pub async fn communicate_certificates<Socket: AsyncWrite + AsyncRead + Unpin + Send, CertApi: CertificateApi + Sync + Send + Clone + 'static>(
     vsock: &mut Socket,
     cert_api: CertApi,
 ) -> Result<(), String> {
-    // Process certificate requests until we get the SetupSuccessful message
-    // indicating that the enclave is done with setup. There can be any number
+    // Process certificate requests. There can be any number
     // of certificate requests, including 0.
     loop {
+        let cert_api = cert_api.clone();
```
tvsfx, line 94: Nit: I wouldn't clone unless necessary.

```diff
diff --git a/vsock-proxy/parent/lib/lib.rs b/vsock-proxy/parent/lib/lib.rs
index c31fb0a..2907c50 100644
--- a/vsock-proxy/parent/lib/lib.rs
+++ b/vsock-proxy/parent/lib/lib.rs
@@ -91,7 +91,6 @@ pub async fn communicate_certificates<Socket: AsyncWrite + AsyncRead + Unpin + S
     // Process certificate requests. There can be any number
     // of certificate requests, including 0.
     loop {
-        let cert_api = cert_api.clone();
 
         let msg: SetupMessages = vsock.read_lv().await?;
 
@@ -100,6 +99,7 @@ pub async fn communicate_certificates<Socket: AsyncWrite + AsyncRead + Unpin + S
                 return Ok(())
             },
             SetupMessages::CSR(csr) => {
+                let cert_api = cert_api.clone();
                 handle_csr_message(vsock, cert_api, csr).await?
             },
             other => {
```

vsock-proxy/parent/lib/lib.rs
```diff
@@ -78,7 +78,7 @@ pub async fn handle_csr_message<Socket: AsyncWrite + AsyncRead + Unpin + Send, C
             // Failures may be silently dropped (i.e., when the enclave renews certificate in a
             // background task periodically. Ensure it can retry after some time and doesn't keep
             // waiting.
-            let _ = vsock.write_lv_bytes(&[]);
+            let _ = vsock.write_lv_bytes(&[]).await;
```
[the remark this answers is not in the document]
raoulstrackx: Yes let's return the error to keep it in line with the Ok result

vsock-proxy/parent/lib/lib.rs
```diff
@@ -44,41 +48,65 @@ fn node_agent_address() -> Option<String> {
...
             Err(e)
         },
     }
 }
 
-pub async fn communicate_certificates<Socket: AsyncWrite + AsyncRead + Unpin + Send, CertApi: CertificateApi>(
+pub async fn communicate_certificates<Socket: AsyncWrite + AsyncRead + Unpin + Send, CertApi: CertificateApi + Sync + Send + Clone + 'static>(
```
tvsfx, line 87: Nit: is the `Sync` bound required here? I don't think you ever capture the `CertApi` argument by reference?

vsock-proxy/parent/lib/lib.rs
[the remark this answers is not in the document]
tvsfx: Yeah, I guess there's something to be said for returning the original error; I imagine that one's more important, but I don't know since it's not my code :)

vsock-proxy/parent/lib/lib.rs
raoulstrackx: This also will be cleaned up in a follow up PR. This can just be a type instead of an instance of a type. This is also out of scope for this PR

PR #56, opened 2025-12-22
[RTE-662] Update shiplift crate to be compatible with newer versions of docker

aditijannu:
We plan to deploy the salmiac converter in an environment where the docker version installed would be at least v 29.1.3. This version of docker uses Docker Engine API v1.52. This upgrade requires a change to the shiplift crate we use to interact with the docker engine.
Corresponding shiplift update PR - https://github.com/fortanix/shiplift/pull/13

Todo: Update Cargo.toml and .lock file once the shiplift PR is merged.


tools/container-converter/src/image.rs
```diff
@@ -55,11 +55,11 @@ impl<'a> ImageWithDetails<'a> {
     }
 
     pub fn user_and_group(&self) -> (User, User) {
-        ImageWithDetails::extract_user_and_group(&self.details.config.user)
+        ImageWithDetails::extract_user_and_group(&self.details.config.user.clone().unwrap_or_default())
     }
 
     pub fn working_dir(&self) -> WorkingDir {
-        WorkingDir::from(self.details.config.working_dir.deref())
+        WorkingDir::from(self.details.config.working_dir.clone().unwrap_or_default().as_str())
```
raoulstrackx: Instead of resorting to the default, could we return an `Option` here instead? Similar comment for the other changes.

aditijannu: The defaults for working_dir is set here - https://github.com/fortanix/salmiac/blob/e1fb0dc812941c10938272e826a2b41554dc98d8/api-model/src/enclave.rs#L93

The defaults for user and group can be set only at run time, it is done in the enclave-startup program here - https://github.com/fortanix/salmiac/blob/e1fb0dc812941c10938272e826a2b41554dc98d8/enclave-startup/src/main.rs#L91

The other fields like comment, author, docker_version etc are not used by salmiac at this point.

tools/container-converter/Cargo.toml
```diff
@@ -24,7 +24,7 @@ openssl = "0.10"
 rand = "0.8.4"
 serde = { version = "1.0.127", features = ["derive"] }
 serde_json =  { git = "https://github.com/fortanix/serde-json.git", branch = "base64_bytes" }
-shiplift = { git = "https://github.com/fortanix/shiplift.git" }
+shiplift = { git = "https://github.com/fortanix/shiplift.git", branch = "aj/RTE-662" }
```
raoulstrackx: With the shiplift PR merged, can you point it back to master?

raoulstrackx (approved):

PR #57, opened 2026-04-16
[RTE-880]: Upgrade mbedtls versions.

sardok:
Tested build within mono-repo.

### DIFF
--- a/tools/enclaveos-encrypted-fs/Cargo.lock
+++ b/tools/enclaveos-encrypted-fs/Cargo.lock
@@ -543,7 +543,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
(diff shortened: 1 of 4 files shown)


dlegalov-fortanix:
@sardok Is there a reason why only the `Cargo.lock` files are updated, and not the `Cargo.toml`?

dlegalov-fortanix:
I ask because the `mbedtls` version is changed here: https://github.com/fortanix/salmiac/pull/57/changes#diff-a4a07410bd7e7a74e40a2024f6d8eb0c9e26cce0574002ec041a7187b9378872R1292
but there is no change to the `.toml` files, such as
in https://github.com/fortanix/salmiac/blob/master/vsock-proxy/enclave/Cargo.toml#L28

Taowyoo (dismissed):

Taowyoo:
I agree. If internal repo have no issue to explicitly upgrade mbedtls to 0.13.5, then it would be good to also update `Cargo.toml`

sardok:
Updated the Cargo.toml files. Thank you.

dlegalov-fortanix (approved):
Looks good to me with the Cargo.toml changes!

PR #58, opened 2026-04-21
[RTE-910] Add script to convert a container image to a CPIO archive

rui-alm:
🔧 Uses [skopeo](https://github.com/containers/skopeo) and [umoci](https://umo.ci/) to extract the container image filesystem.

📦 Uses `cpio` to create the CPIO archive in New ASCII format.

🧪 Tested using a container image containing a basic init script which mounted a rootfs based on nginx image and handed over to `init.c`, using a generic Ubuntu kernel on QEMU.

### DIFF
--- a/docker/enclave-base/container_image_to_cpio.sh
+++ b/docker/enclave-base/container_image_to_cpio.sh
@@ -8,14 +8,13 @@ function show_usage() {
         [-h]                Show help
         [-k]                Keep intermediary files (default: false)
         [-o output_file]    CPIO output file (default: stdout)
-        -i <oci_url>        Image URL supported by skopeo copy (containers-storage, dir, docker, docker-archive, docker-daemon, oci, oci-archive, ostree, sif, tarball)
-        -t <image_tag>      Image tag"
+        -i <oci_url>        Image URL supported by skopeo copy (containers-storage, dir, docker, docker-archive, docker-daemon, oci, oci-archive, ostree, sif, tarball)"
 }
 
 keep_files=0
 output_file=/dev/stdout
-tag=latest
-while getopts "hi:ko:t:" option; do
+tag=salmiac
+while getopts "hi:ko:" option; do
     case $option in
         h)
             echo "Convert a container image to a CPIO New ASCII format"
@@ -31,9 +30,6 @@ while getopts "hi:ko:t:" option; do
         o)
             output_file=$(realpath "$OPTARG")
             ;;
-        t)
-            tag="$OPTARG"
-            ;;
         *)
             show_usage
             exit 1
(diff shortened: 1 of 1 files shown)


docker/enclave-base/container_image_to_cpio.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# Extracts filesystem layers using `skopeo`, extracts the filesystem using `umoci` and builds the CPIO archive using `cpio`.
+
+function show_usage() {
+    echo "Usage: $0 [OPTION...]
+        [-h]                Show help
+        [-k]                Keep intermediary files (default: false)
+        [-o output_file]    CPIO output file (default: stdout)
+        -i <oci_url>        Image URL supported by skopeo copy (containers-storage, dir, docker, docker-archive, docker-daemon, oci, oci-archive, ostree, sif, tarball)
+        -t <image_tag>      Image tag"
```
jethrogb: I think tag should be specified as part of `-i`. You can use a fixed tag between skopeo and umoci.

docker/enclave-base/container_image_to_cpio.sh
```diff
@@ -0,0 +1,70 @@
...
+    esac
+done
+
+if [ -z "${oci_url}" ]; then
+    show_usage
+    exit 1
+fi
+
+# Create temporary directory to extract image layers
+skopeo_tmp_dir=$(mktemp --directory --tmpdir="${PWD}")
```
jethrogb:
```suggestion
skopeo_tmp_dir="$(mktemp --directory --tmpdir="${PWD}")"
```
Same thing below.

docker/enclave-base/container_image_to_cpio.sh
rui-alm: I've removed the tag, used a shared tag "salmiac" and escaped the temporary paths: 5f23a9aa0c49928bf7d362967c93e9cca0190d95

jethrogb (dismissed):

docker/enclave-base/container_image_to_cpio.sh
```diff
@@ -0,0 +1,66 @@
...
+# Copy image filesystem layers
+echo skopeo copy "${oci_url}" "oci:${skopeo_tmp_dir}:${tag}" >&2
+skopeo copy "${oci_url}" "oci:${skopeo_tmp_dir}:${tag}" >&2
+
+# Create temporary directory to unpack rootfs
+umoci_tmp_dir="$(mktemp --directory --tmpdir="${PWD}")"
+
+# Extract image into a bundle
+echo umoci unpack --rootless --image "${skopeo_tmp_dir}:${tag}" "${umoci_tmp_dir}" >&2
+umoci unpack --rootless --image "${skopeo_tmp_dir}:${tag}" "${umoci_tmp_dir}"
```
tvsfx: Two nits;
1. Should this print to stderr for symmetry with other commands?
2. Why dont you do `set -x; umoci unpack --rootless --image "${skopeo_tmp_dir}:${tag}" "${umoci_tmp_dir}"; set +x`? (or just `set -x` after argument parsing above)?

tvsfx (dismissed):

docker/enclave-base/container_image_to_cpio.sh
rui-alm: umoci is already printing to stderr, but I guess it doesn't hurt to ensure any stdout is redirected. 80bd8fe8245926d85ca2b4292bcb2abff14114d6

rui-alm:
We are currently considering invoking these commands directly from the converter, instead of invoking this script. We can use this script as a reference it may not be necessary to merge this PR.

aditijannu (approved):

PR #59, opened 2026-04-21
Upgrade sys-mount to 3.1.0

rui-alm:
Depending on the clang version, the old `bindgen` might fail: https://github.com/rust-lang/rust-bindgen/issues/2312.

This PR causes the `bindgen` version to be bumped to a version that fixes the issue. For newer clang versions this PR and #57 (which updates mbedtls, bumping bindgen as well) fix salmiac's build.


tvsfx (approved):

MihirLuthra:
I was trying sys-mount 3.1 locally with the converter and noticed there was a fix in sys-mount 3.1: https://github.com/pop-os/sys-mount/pull/36 that effects us. I suppose we would need to add `explicit_loopback()` to our builder as well?

PR #60, opened 2026-04-23
[RTE-914] Add parent-base-snp image for SEV-SNP host

anubhavg64:
Adds a new docker/parent-base-snp/ directory with a Dockerfile, dnsmasq.d config, and a  
tag file. This is a side-by-side SEV-SNP variant of the existing docker/parent-base/
(Nitro) and does not modify the Nitro path.                                           
                                                                                           
The Dockerfile produces a host-side image that packages QEMU and Salmiac's standard host 
tooling, which together are what will be used to launch SEV-SNP confidential VMs.
                                                                                           
Verified locally that docker build completes successfully and the resulting image runs   
cleanly.


docker/parent-base-snp/dnsmasq.d
```diff
@@ -0,0 +1,23 @@
+# If you want dnsmasq to listen for DHCP and DNS requests only on
```
aditijannu: This is an exact copy of what we have for nitro parent-base image, we can reuse the existing file.

aditijannu:
You can update `.github/workflows/build-docker-images.yml` to include testing of parent-base-snp image build.

docker/parent-base-snp/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+        libslirp-dev \
+        meson \
+        ninja-build \
+        pkg-config \
+        python3 \
+        python3-venv \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN git clone --depth 1 --branch "${QEMU_VERSION}" \
+        https://gitlab.com/qemu-project/qemu.git /src/qemu
```
jethrogb: Please download from https://download.qemu.org/qemu-10.2.2.tar.xz and check the hash after downloading.

docker/parent-base-snp/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+        iptables \
+        jq \
+        kmod \
+        libaio1t64 \
+        libcap-ng0 \
+        libglib2.0-0t64 \
+        libpcap-dev \
+        libpixman-1-0 \
+        libseccomp2 \
+        libslirp0 \
```
jethrogb, line 113: Can you please list explicitly for each lib why it's needed? (I assume QEMU dependency)

docker/parent-base-snp/Dockerfile
anubhavg64: okay, will switch from git clone to the official tarball with SHA-256 verification.

docker/parent-base-snp/Dockerfile
anubhavg64: Your assumption is almost right about half the list is QEMU runtime deps, the rest is Salmiac networking / utilities. Adding their purpose below:

```
QEMU runtime libraries, linked by the binaries built in stage 1:
libaio1t64      - libaio (async disk I/O)                                        
libcap-ng0      - libcap-ng (POSIX capability drops for sandboxing)
libglib2.0-0t64 - GLib (QEMU's core utility library)                             
libpixman-1-0   - pixman (guest framebuffer pixel ops; still linked even with --disable-gtk/sdl/vnc)                        
libseccomp2     - seccomp BPF filter (--enable-seccomp)
libslirp0       - libslirp (user-mode networking, --enable-slirp)                
zlib1g          - zlib (compressed disk images, bzImage loading)                 

Host <-> CVM networking for the fortanix-tap0 interface:                           
 dnsmasq, iproute2, iptables, net-tools                                           

Backing storage for the CVM's persistent disk (same as Nitro):                     
  nbd-server                                                                                                                                      

Kernel-module loader for KVM / VFIO setup on the host:                             
  kmod                                              

General userspace utilities used by Salmiac tooling:
  ca-certificates, curl, jq, sudo, systemctl                                       

Diagnostics:                                                                       
  strace, tcpdump
```

jethrogb: Thanks for the overview. I intended for this to be in a comment in the file

anubhavg64: Sure, will add this as a comment.

.github/workflows/build-docker-images.yml
```diff
@@ -14,6 +15,7 @@ on:
     paths:
       - 'docker/enclave-base/**'
```
aditijannu: I see that this workflow is still not triggered, might be a good idea to update the paths to `docker/**`

aditijannu (dismissed):

.github/workflows/build-docker-images.yml
anubhavg64: The "Build and publish Docker Images" workflow did not run because of two underlying issues.
First, it only triggers when a PR is opened, not on synchronize (when new commits are pushed to it), so any updated PR is silently skipped. 
Second, the path filter lists each image directory by name, which means adding a new variant like parent-base-snp needs a workflow edit,and that edit doesn't help retroactively for the original PR open. 
Proposing two small fixes, also re-trigger on push, and broaden the path filter to docker/** so future image variants pick up automatically.
Will fold both into this unless there's a reason to split.

aditijannu (approved):

docker/parent-base-snp/Dockerfile
```diff
@@ -0,0 +1,130 @@
...
+        systemctl \
+        tcpdump \
+        zlib1g \
+    && rm -rf /var/lib/apt/lists/*
+
+# QEMU built above.
+COPY --from=qemu-builder /opt/qemu /opt/qemu
+ENV PATH="/opt/qemu/bin:${PATH}"
+
+COPY --from=parent-base dnsmasq.d /etc/dnsmasq.d
```
MihirLuthra, line 127: I suppose here we meant to copy `dnsmasq.d` from the host and not from `parent-base`?

PR #61, opened 2026-04-24
RTE-922: Refactor vsock-proxy for multi-platform support

MihirLuthra:
This PR refactors `vsock-proxy` to put platform specific code behind a cfg.
Essentially, both parent and enclave crates have a new module `platform`, with structure:
```
platform
├── mod.rs
└── nitro.rs
```
and conditionally compiled with:

```
#[cfg(platform = "<platform_name>")]
```


MihirLuthra:
I haven't tested things here with the new changes. Will update here once done.
Also for compiling converter, this change needs to go in first: https://github.com/fortanix/salmiac/pull/59

vsock-proxy/enclave/nitro/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "vsock-proxy-enclave-nitro"
+version = "0.1.0"
+edition = "2024"
```
rui-alm: The CI build is failing. I guess our toolchain is expected to be `2018`.

rui-alm (changes requested):

vsock-proxy/parent/nitro/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "vsock-proxy-parent-nitro"
+version = "0.1.0"
+edition = "2024"
```
dlegalov-fortanix: This may also break CI due to the the old version needed.

vsock-proxy/enclave/nitro/Cargo.toml
MihirLuthra: Oh, didn't notice the edition 🙂. I have updated this PR to not have new crates anymore (and rely on cfgs instead), so this shouldn't be an issue anymore.

tools/container-converter/build.rs
```diff
@@ -8,6 +8,10 @@ use std::{env, fs};
 // to avoid problems runtime linking errors with libnss
 const ENCLAVE_STARTUP_TARGET: &str = "x86_64-unknown-linux-musl";
 
+const VSOCK_PROXY_PLATFORM_ENV: &str = "VSOCK_PROXY_PLATFORM";
```
rui-alm: Should we change these variables to something more generic, like `SALMIAC_PLATFORM`? We have the converter that needs to be split into multi platforms as well and we should be able to reuse this configuration.

MihirLuthra: Yes, `SALMIAC_PLATFORM` sounds like a good option. Will update

aditijannu: ENCLAVEOS prefix is better suited here than SALMIAC, please add a relevant documentation here - https://github.com/fortanix/salmiac/blob/master/ENV_VARS.md

jethrogb: Why? This is an internal build variable. It's not immediately clear to me why we'd use a product name.

MihirLuthra:
TODO: Add README

tools/container-converter/build.rs
```diff
@@ -67,6 +73,8 @@ fn main() -> Result<(), Box<dyn Error>> {
     .status()
     .expect("Failed to build enclave-startup project");
 
+    assert!(status.success(), "Failed to build enclave-startup project: {}", status);
```
tvsfx: nit: no capital on assert/expect

vsock-proxy/parent/src/platform/nitro.rs
```diff
@@ -0,0 +1,84 @@
...
+
+fn env_var_or_default<T: ToString>(var_name: &str, default: T) -> String {
+    env::var(var_name).unwrap_or_else(|_| default.to_string())
+}
+
+pub(crate) fn enclaveos_debug_enabled() -> bool {
+    env::var(ENCLAVEOS_DEBUG_ENV).as_deref() == Ok(ENCLAVEOS_DEBUG_VALUE)
+}
+
+pub(crate) fn enable_client_log_forwarding() -> bool {
```
tvsfx: nit: this sounds like an action to me, rather than a pure, boolean function

vsock-proxy/parent/src/platform/nitro.rs
```diff
@@ -0,0 +1,84 @@
...
+
+const NITRO_CLI: &str = "nitro-cli";
+const ENCLAVE_NAME: &str = "enclave";
+const EIF_PATH: &str = "/opt/fortanix/enclave-os/enclave.eif";
+
+fn env_var_or_default<T: ToString>(var_name: &str, default: T) -> String {
+    env::var(var_name).unwrap_or_else(|_| default.to_string())
+}
+
+pub(crate) fn enclaveos_debug_enabled() -> bool {
```
tvsfx: make private?

tvsfx (commented):
Looks good to me! :) Just a few nits.

build-support/src/lib.rs
rui-alm, line 1: Since we'll need this for the converter and api-model, should this be moved out of vsock proxy? Perhaps a build-support crate that we can include on the other projects as a build dependency so that they can pickup the platform using the same environment variable.

tools/container-converter/build.rs
MihirLuthra: I have named it `SALMIAC_PLATFORM`, as it is a build time env. Also, I see that `ENV_VARS.md` currently has documentation for runtime variables. I have instead added docs in separate `README` for this case.

vsock-proxy/parent/src/platform/nitro.rs
MihirLuthra: Thanks for the review! I have addressed the comments.

build-support/src/lib.rs
MihirLuthra: Yes, makes sense. I have moved the build-support to workspace root instead.

vsock-proxy/parent/src/platform/nitro.rs
tvsfx: You accidentally made this public instead of private :sweat_smile:

tvsfx (dismissed):

vsock-proxy/parent/src/platform/nitro.rs
MihirLuthra: oops 🙂. Fixed it now

tvsfx (approved):

rui-alm (approved):

PR #62, opened 2026-04-28
[RTE-915] Add enclave-base-snp image for SEV-SNP guest

anubhavg64:
The image stays minimal, same base set as the Nitro enclave-base (ca-certificates, cryptsetup, nbd-client) plus the userspace NVIDIA driver packages (libnvidia-compute-595, nvidia-utils-595) needed for guest-side GPU access.
Userspace is pinned to the 595 branch to match the version being used on the kernel-modules side.
Kernel modules themselves are not installed here, they're handled separately and shipped with the guest kernel.


docker/enclave-base-snp-gpu/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+#                       libnvidia-ptxjitcompiler.so.1
+#     nvidia-utils-${NVIDIA_DRIVER_BRANCH}
+#                     - nvidia-smi, libnvidia-ml.so.1, libnvidia-cfg.so.1
+RUN apt-get update \
+    && apt-get upgrade -y \
+    && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        cryptsetup \
+        nbd-client \
+        libnvidia-compute-${NVIDIA_DRIVER_BRANCH} \
```
tvsfx, line 36: How did you select these packages to install?

- What's the difference between https://packages.ubuntu.com/questing/nvidia-headless-no-dkms-590-server-open and libnvidia-compute-${NVIDIA_DRIVER_BRANCH}? Do we need any of the other libs in the headless package?
- Do we need `nvidia-smi` in salmiac, or why do we install `nvidia-utils`?
- I think using NVIDIA_DRIVER_BRANCH=595 is fine if we decide to go for 24.04 (which I think we will? just checking) - I did notice yesterday that the 595 package does not exist for every Ubuntu release yet.

docker/enclave-base-snp-gpu/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+        nvidia-utils-${NVIDIA_DRIVER_BRANCH} \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# Fail-fast sanity checks. Catch kernel-module leakage from apt
+# dependencies, and verify the two artifacts the workload actually needs.
+RUN if find /lib/modules -name 'nvidia*.ko*' 2>/dev/null | grep -q .; then \
```
tvsfx, line 43: Nice, this is good :)

docker/enclave-base-snp-gpu/tag
tvsfx, line 1: Do we use this `tag` file anywhere? I'm probably missing something

anubhavg64: Fair point, the tag files aren't read by anything today. Plan is to use it once we update build-conv-container.sh to read versions from these files instead of the hardcoded ones. Keeping it here so the new image has a tag from the start.

docker/enclave-base-snp-gpu/Dockerfile
anubhavg64: 1. I picked the minimum userspace set from first principles (libcuda.so + nvidia-smi/NVML), not from a meta-package. Looked at nvidia-headless-no-dkms-595-server-open, looks like it pulls in nvidia-kernel-source-595-server-open (few extra MBs of driver source), which we don't want here since the kernel modules would be built-in. 
We could switch to the -server variants of the individual packages, happy to do that if you really want that way?

3. nvidia-utils-* brings: nvidia-smi CLI and libnvidia-ml.so.1 (NVML). 
We need NVML for any programmatic GPU queries from the guest agent. I'd lean keeping it for operator/debug visibility too.
Also FYI, we had hit this issue while running the attestation client in a CVM, without this we get the following error.
`libnvidia-ml.so: cannot open shared object file: No such file or directory`

4. Good point, honestly I just went with this considering we are having same for Nitro images.

docker/enclave-base-snp-gpu/Dockerfile
tvsfx, line 1: Can we rename this to clarify that this is the image with GPU support?

docker/enclave-base-snp-gpu/Dockerfile
tvsfx: >  libnvidia-ml.so: cannot open shared object file: No such file or directory

Yes, presumably this is the only library we have confirmed that we actually need; I don't think we want/need `nvidia-smi` in our hardened prod image?
Can you put a TODO here with a link to the userspace hardening ticket?

docker/enclave-base-snp-gpu/Dockerfile
tvsfx: We should probably also clarify somewhere that `enclave-base` is suitable both for AWS nitro and SNP without GPU support.

docker/enclave-base-snp-gpu/Dockerfile
anubhavg64: TODO: RTE-944 Userspace hardening

tvsfx: I meant; add the TODO here in the code, so people can find it. No need for an (internal!) URL, just put the ticket #

tvsfx (approved):

PR #64, opened 2026-05-06
RTE-919 - Converter SNP API and split

rui-alm:
:keyboard: Implements SNP API from RTE-907.

:scissors: Splits nitro specific code from common and SNP specific code.

Depends on #61


api-model/src/lib.rs
```diff
@@ -65,8 +71,9 @@ impl<'de> Deserialize<'de> for HexString {
             }
 
             fn visit_str<E: Error>(self, string: &str) -> Result<HexString, E> {
-                Ok(HexString::from_str(string)
-                    .map_err(|_| Error::invalid_value(serde::de::Unexpected::Str(string), &"hex encoded string"))?)
+                Ok(HexString::from_str(string).map_err(|_| {
```
tvsfx: Can you do the formatting in a separate PR? Rust-analyzer supports LSP's Range Formatting, so you can highlight and format only the parts you've worked on in most modern editors.

rui-alm: I've spent a few hours trying to get range formatting to work and it doesn't. `rustfmt` formats more than it's being told to, even when I invoke it from the command line with specific lines to format.

For now I'll start reverting the formatting changes, but if you have a rust-analyzer/rustfmt configuration I can use it would be appreciated. neovim or lazyvim configuration would be even better.

rui-alm: e20183035a93833211ddd561b96e141dad4c860b

api-model/Cargo.toml
```diff
@@ -11,3 +11,8 @@ repository = "https://github.com/fortanix/salmiac"
 hex = "0.4"
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_json = { git = "https://github.com/fortanix/serde-json.git", branch = "base64_bytes" }
+
+[lints.rust]
```
tvsfx, line 21: ~~Since this crate always depends on another crate setting the `platform` config, it might make sense to have `build-support` offer a function that _only_ checks config rather than setting + checking, so we can use it in a `build.rs` here rather than having to hard-code the platform values.~~ Nonsense, see below.

I think we want this regardless, since the `println!("cargo::rerun-if-env-changed={PLATFORM_ENV}");` line does not affect dependencies, meaning we want it here too.

tools/container-converter/src/image_builder/nitro/enclave.rs
```diff
@@ -0,0 +1,477 @@
...
+            create_nitro_image(&result.image.reference, &nitro_image_path).await?
+        };
+
+        info!("Nitro image has been created!");
+
+        Ok(nitro_measurements)
+    }
+}
+
+#[cfg(test)]
```
tvsfx: Based on the `use` statements below, it looks like these tests aren't actually platform-specific;

* The pure `ArchiveSize` tests should've been placed with this struct
* The `EnclaveImageBuilder` tests are run on the platform-indepent `EnclaveImageBuilder`, not the one defined here.

tools/container-converter/src/nitro.rs
```diff
@@ -0,0 +1,225 @@
+/* Copyright (c) Fortanix, Inc.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#[cfg(platform = "nitro")]
```
tvsfx: Unlikely that the `cfg`s here and below (in snp.rs) are necessary, given that the file is named `nitro` and should hence be configed out in its entirety on different platforms.

tools/container-converter/src/main.rs
```diff
@@ -4,47 +4,51 @@
...
 
-use std::fs;
-
-use api_model::converter::NitroEnclavesConversionRequest;
 use clap::{App, AppSettings, Arg, ArgMatches};
 use env_logger;
-use log::error;
+use std::fs;
+
+#[cfg(platform = "nitro")]
```
tvsfx, line 11: Can't you just have one main function with something like:

```rust
#[cfg(platform = "nitro")]
use container_converter::nitro as platform;
#[cfg(platform = "snp")]
use container_converter::snp as platform;

#[tokio::main]
async fn main() -> Result<(), String> {
    ...
    let console_arguments = console_arguments(platform::PLATFORM_NAME), // or just change value inline
    );
    ...
    platform::process_request(&request_file).await
}
```

It's imo even better to handle this selective exporting in `container_converter/mod.rs` (using e.g.`#[cfg(platform = "nitro")] pub(crate) use nitro::*` so all non-private stuff gets exported)

tools/container-converter/src/nitro.rs
```diff
@@ -0,0 +1,225 @@
...
+};
+use api_model::{
+    enclave::UserConfig,
+    nitro::{NitroEnclavesConversionRequest, NitroEnclavesConversionResponse},
+};
+use log::{debug, error, info};
+use shiplift::Docker;
+use tempfile::TempDir;
+
+pub async fn process_request(request_file: &str) -> std::result::Result<(), String> {
```
tvsfx: No need to duplicate `process_request`/`run` and all of  `run0` ( I have the impression that the only logic that differs is the `measurement` you return?); you can play with exports as outlined above.

tools/container-converter/Cargo.toml
```diff
@@ -34,3 +34,8 @@ url = "2.2.2"
 
 [dev-dependencies]
 chrono = "0.4.22"
+
+[lints.rust]
```
tvsfx, line 41: I dont think you need to repeat `check-cfg` again; the build file should take care of which.
You do need to add platform support for `container-converter` itself to its `build.rs`

api-model/Cargo.toml
```diff
@@ -11,3 +11,8 @@ repository = "https://github.com/fortanix/salmiac"
 hex = "0.4"
 serde = { version = "1.0", features = ["derive"], optional = true }
 serde_json = { git = "https://github.com/fortanix/serde-json.git", branch = "base64_bytes" }
+
```
sardok, line 14: I was going to suggest the following change but realized that mono-repo still uses rustc 1.57. I think for the following 1.60 is required at least.

```
[features]
serde = ["dep:serde"]
```

api-model/src/snp.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+#[derive(Clone, Eq, PartialEq, Debug)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct SNPEnclavesConfig {
+    /// SNPEnclaves measurements of the converted image
+    pub measurements: SNPEnclavesMeasurements,
+}
+
+//#[derive(Clone, Eq, PartialEq, Debug, Hash)]
```
tvsfx: remove comment

api-model/Cargo.toml
sardok: Ok, as i confirmed with @aditijannu this project won't be compiled with 2021 compiler. So i would suggest the above change. Explicit features provide more clarity over implicit optional dep features.

api-model/Cargo.toml
```diff
@@ -11,3 +11,8 @@ repository = "https://github.com/fortanix/salmiac"
...
+[lints.rust]
+unexpected_cfgs = { level = "deny", check-cfg = [
+  'cfg(platform, values("nitro", "snp"))',
+] }
```
sardok: I wonder why we decided to go `platform` variable for `nitro` & `snp` as opposed to `nitro` & `snp` features? One reason could be features are aditive, `platform` variable is mutually exclusive. I don't see why `nitro` & `snp` shouldn't be additive if this is the reason. Because, probably, we will want to be able to convert both `nitro` & `snp` images in the backend, therefore we would need them both, so introducing `platform` variable would cause a problem in the backend if i am not missing something.

tools/container-converter/Cargo.toml
```diff
@@ -34,3 +34,8 @@ url = "2.2.2"
...
+unexpected_cfgs = { level = "deny", check-cfg = [
+  'cfg(platform, values("nitro", "snp"))',
+] }
```
sardok: Same comment: https://github.com/fortanix/salmiac/pull/64/changes#r3202370135

api-model/Cargo.toml
sardok: Also, in order to compile with custom `cfg` you would need to do something like `RUSTFLAGS='--cfg platform="snp"' cargo build`. That is fine, but in CI scripts and or local development, it easier to build/test with features such as `cargo build --all-features`.

tools/container-converter/src/lib.rs
```diff
@@ -3,6 +3,11 @@
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#[cfg(platform = "nitro")]
```
tvsfx, line 6: Haven't really thought this through, but another way to structure this would be folders `enclave` and `parent`, and then config-specific `pub use` statements in each one of them (similar to what I proposed for `main` below). That would allow identical paths for identically-named items in e.g. `nitro/enclave.rs` and `snp/enclave.rs`, rather than having `nitro::enclave` and `snp::enclave`.

api-model/Cargo.toml
tvsfx: The reason is indeed additiveness; we don't want one binary that supports more than one platform, for hardening reasons.

tools/container-converter/src/image_builder/enclave/nitro.rs
```diff
@@ -0,0 +1,477 @@
...
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use std::path::Path;
+use std::sync::mpsc::Sender;
+
+use api_model::enclave::{EnclaveManifest, UserConfig};
+use docker_image_reference::Reference as DockerReference;
```
sardok, line 11: Need `use log::info` here.

api-model/Cargo.toml
tvsfx: https://github.com/fortanix/salmiac/pull/61/changes#diff-ee17749c4322a922264bf6d8624a247193e14c7ce6c295b3eface9aeef0b84bf
Mihir set up a new crate to make building this nicer; there's no need for `RUSTFLAGS`

tools/container-converter/src/lib.rs
```diff
@@ -3,6 +3,11 @@
...
+pub mod nitro;
+#[cfg(platform = "snp")]
+pub mod snp;
+
```
sardok, line 10: Many imports are not used. They should be removed.

tools/container-converter/src/nitro.rs
sardok: This shouldn't be needed as this file is included in this condition already.

tools/container-converter/src/nitro.rs
```diff
@@ -0,0 +1,225 @@
...
+    get_enclave_base_image, get_parent_base_image,
+    image::{docker_reference, output_docker_reference, ImageKind, ImageToClean, ImageWithDetails},
+    image_builder::{
+        enclave::{get_image_env, EnclaveImageBuilder, EnclaveSettings},
+        parent::ParentImageBuilder,
+    },
+    preserve_images_list, push_result_image, validate_request, ConverterError, ConverterErrorKind,
+    Result, ENCLAVE_IMAGE, PARENT_IMAGE,
+};
+#[cfg(platform = "nitro")]
```
sardok: Same as above comment.

tools/container-converter/src/image_builder/parent/nitro.rs
```diff
@@ -0,0 +1,204 @@
...
+                message,
+                kind: ConverterErrorKind::ParentImageCreation,
+            })?;
+
+        info!("Parent image has been created!");
+
+        Ok(result)
+    }
+
+    fn create_requisites(
```
tvsfx, line 87: Presumably, figuring out which of the below is nitro-specific is left as a task for the implementor of RTE-941?

tools/container-converter/src/image_builder/parent/nitro.rs
```diff
@@ -0,0 +1,204 @@
...
+            parent_bin.display()
+        )
+    }
+
+    fn cpu_count_env_var(&self) -> String {
+        format!(
+            "CPU_COUNT={}",
+            self.start_options
+                .cpu_count
+                .unwrap_or(crate::image_builder::parent::ParentImageBuilder::DEFAULT_CPU_COUNT)
```
tvsfx, line 190: Even though this is likely to change in the future, it's a little odd to me that `DEFAULT_CPU_COUNT` is not platform specific, but the function to read the CPU count (the only location where it's used) is.

tools/container-converter/src/image_builder/parent/snp.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use api_model::snp::SNPEnclavesConversionRequestOptions;
+use tempfile::TempDir;
+
+use crate::{image::ImageWithDetails, Result};
+
+pub(crate) struct ParentImageBuilder<'a> {
+    pub(crate) parent_image: String,
+    pub(crate) dir: &'a TempDir,
+    pub(crate) start_options: SNPEnclavesConversionRequestOptions,
```
tvsfx, line 13: nit (not important right now): follow `nitro` and wrap the common `ParentImageBuilder` here

api-model/Cargo.toml
sardok: > The reason is indeed additiveness; we don't want one binary that supports more than one platform, for hardening reasons.

What is hardening reasons? I wonder how are we planning to support two platforms in the backend.

Also one thing i don't like with `cfg` is, they are implicitly propagated. So if you have project which depends on `dep1`, instead explicitly stating `dep1:feature`, you're implicitly relying on `dep1` on this particular `cfg`.

sardok: Ok, @tvsfx explained the rational behind `cfg` in private chat. I'm ok with it.

tools/container-converter/src/image_builder/enclave/mod.rs
```diff
@@ -130,17 +95,15 @@ pub(crate) struct EnclaveImageBuilder<'a> {
 }
 
 impl<'a> EnclaveImageBuilder<'a> {
```
sardok, line 115: Now `create_manifest_file`, `create_block_file` and `enclave_image` of `EnclaveImageBuilder` must be `pub(crate)`.

tools/container-converter/src/image_builder/enclave/nitro.rs
```diff
@@ -0,0 +1,477 @@
...
+            user_config,
+            file_system_config,
+            is_debug,
+            env_vars,
+            enable_overlay_filesystem_persistence,
+            ccm_backend_url,
+            dsm_configuration,
+        };
+
+        crate::image_builder::enclave::EnclaveImageBuilder::create_manifest_file(
```
sardok, line 126: This should be `self.enclave_image_builder.create_manifest_file`.

api-model/Cargo.toml
MihirLuthra: Interesting, @tvsfx do you mean there is case where we want only emit `rustc-check-cfg` and not set `rustc-cfg=platform=\"{}\"`? Seeing in the api-model crate, I see there is usage of `#[cfg(platform = "...")]`.

MihirLuthra: It might also be nice if later we refactor all Salmiac to be under a single workspace. Then, we can have either:
```
[workspace.lints.rust]
warnings = "deny"
```
or
```
[workspace.lints.rust]
unexpected_cfgs = { level = "deny" }
```

tvsfx: > Interesting, @tvsfx do you mean there is case where we want only emit rustc-check-cfg and not set rustc-cfg=platform=\"{}\"? Seeing in the api-model crate, I see there is usage of #[cfg(platform = "...")].

Turns out I misremembered how custom `cfg` options work; I thought that if you set `cfg` for a crate then all of its dependents get this `cfg`. I just tried that and it turns out to be plainly untrue :stuck_out_tongue: . That's good actually, I always disliked how I *thought* custom cfg differed from setting features, but in actually they work the same way.

tvsfx: This means that this part is still relevant:

>  we can use it in a build.rs here rather than having to hard-code the platform values

we need a `build.rs` file for `api-model` too

tools/container-converter/src/main.rs
sardok: Also probably it would be nice to add `compiler_error!` when a `platform` is not selected.

api-model/src/lib.rs
rui-alm: I've reverted too much on the previous commit. Additional fixes in bd128499f19efb35d1206c6fb6ca20c115a1d06a

api-model/Cargo.toml
rui-alm: 1582d3621b96757edcc4c87e9839f6d1f0e40fd1

api-model/Cargo.toml
rui-alm: Now that I've rebased the changes from #61 and added f1f207c5bcaf1bb763564c1ee07463d37b4de81b, it's safe to remove the `check-cfg`, since the `build.rs` will fail if we configure any other platform, right?
Otherwise I'll have to revert 7dc028927a30ea8ba7618770e3bd57fa45eb216a.

tools/container-converter/src/image_builder/nitro/enclave.rs
rui-alm: These tests were meant to be moved to the common enclave.rs, but were accidentally left behind. I've removed them in d4955be49ad97d9dbe8df8467815eed2f7ee833d.

Is this what you were referring to, or do you have further suggestions to move these tests?

tools/container-converter/src/nitro.rs
rui-alm: b0c872034b8d092952f122d5d0f088751638a6d5

tools/container-converter/src/main.rs
rui-alm: Beautiful suggestion @tvsfx. Hopefully b0c872034b8d092952f122d5d0f088751638a6d5 is closer to ideal. Any suggestions on static formatting of strings? The `about` seems to expect a static string and I didn't find a pretty way of dealing with it, so I duplicated most of the string between platforms at the moment.

@sardok The compilation fails if no platform is selected, with a message `failed to read SALMIAC_PLATFORM: environment variable not found`.

Would this be enough? I don't think `compiler_error!` would reach main.rs, since the failure is triggered by `build.rs`.

tools/container-converter/src/nitro.rs
rui-alm: b0c872034b8d092952f122d5d0f088751638a6d5

tools/container-converter/Cargo.toml
rui-alm: Do we need any check-cfg on `Cargo.toml` files now? I've removed them on 7dc028927a30ea8ba7618770e3bd57fa45eb216a

api-model/src/snp.rs
rui-alm: ccbe22f2d3ef55d81da68a8e7ea005e1ddb82ae6

tools/container-converter/src/image_builder/enclave/nitro.rs
rui-alm: Thank you, this helped me realize my IDE configurations are broken once again.  bd128499f19efb35d1206c6fb6ca20c115a1d06a

tools/container-converter/src/lib.rs
rui-alm: 05027f21d6cd3160f535c8195f8f5956681c9428

tools/container-converter/src/nitro.rs
rui-alm: b0c872034b8d092952f122d5d0f088751638a6d5

tools/container-converter/src/nitro.rs
rui-alm: b0c872034b8d092952f122d5d0f088751638a6d5

api-model/Cargo.toml
tvsfx: That's correct, `7dc0289` lgtm.

tools/container-converter/src/image_builder/parent/nitro.rs
rui-alm: Yes, there was too much code being moved around already on the PR, and the builders will have most of the platform specific logic, so I thought it would be better to do the split on RTE-941.

@sardok Are you fine with the image_builder split being done as part of RTE-941?

tools/container-converter/src/image_builder/parent/snp.rs
rui-alm: b0c872034b8d092952f122d5d0f088751638a6d5

tools/container-converter/src/image_builder/enclave/mod.rs
rui-alm: bd128499f19efb35d1206c6fb6ca20c115a1d06a

tools/container-converter/src/image_builder/nitro/enclave.rs
tvsfx: Are these tests now gone or is the github UI tricking me? I cant seem to find them anymore

tools/container-converter/src/image_builder/enclave/nitro.rs
rui-alm: The function isn't using any parameters from `self`. Any reason why your suggestion is better? Is it just about style or is there a reason which escapes me?

tools/container-converter/src/main.rs
sardok: > Would this be enough? I don't think compiler_error! would reach main.rs, since the failure is triggered by build.rs.

Yes, thats fine. i was unaware of all these build scripts things.

tools/container-converter/src/image_builder/nitro/enclave.rs
rui-alm: The tests are on `tools/container-converter/src/image_builder/enclave.rs`. I noticed they weren't nitro specific, so I "moved" them to the common enclave.rs, but actually copied them and left them in nitro.rs as well. d4955be49ad97d9dbe8df8467815eed2f7ee833d has removed the duplicates in nitro.rs.

rui-alm:
@MihirLuthra I've reverted some of your changes to add SNP to the GitHub workflows in 56b93a4160d3e8ecbe195532f72c585285c8cd3e.
I've opted for building and testing nitro, and then building and testing SNP. I think this way I'm not triggering unnecessary recompilations due to changes to `SALMIAC_PLATFORM`. I'm not sure is this the best way, so everyone's suggestions are welcome.

tools/container-converter/src/main.rs
tvsfx: @rui-alm 
Since you just need something `'static`, not actually something `const`, you have a bit more freedom. You can just do 
`static MSG: LazyLock<String> = LazyLock::new( || format!(<expression_using_consts>)))` and you can then take static references to this. 

<details>
<summary>Alternative</summary>

Another approach if you want something "const" (not necessary here), is to use `concat!` from `std` inside a macro like
```rust
macro_rules! build_message {
    ($suffix: literal) => {concat!("the custom message is:", $suffix)};
}
```
 if you want to concatenate inline (like `#[cfg(platform = 'snp')] build_message!("snp")`).
If you want to use const identifiers on top of literals to do it (which is a bit more clean), there's a crate for that, see these examples: https://docs.rs/const_format/latest/const_format/macro.formatcp.html#examples
</details>

rui-alm: 94812955d0d3d1b3d6b01dbb464a77eea767af7b

tools/container-converter/src/lib.rs
rui-alm: Something like this? 84ffef542575463ef02bf8533acf5e5f4b0c0ccc

I still haven't attempted to move potential common code between nitro and snp, as it may make more sense to do as part of RTE-941.

tools/container-converter/src/lib.rs
```diff
@@ -3,33 +3,57 @@
...
+#[cfg(platform = "nitro")]
+use crate::image_builder::enclave::nitro::EnclaveImageBuilder as PlatformEnclaveImageBuilder;
+#[cfg(platform = "nitro")]
+use crate::image_builder::parent::nitro::ParentImageBuilder as PlatformParentImageBuilder;
+#[cfg(platform = "nitro")]
+use crate::nitro::create_response;
+#[cfg(platform = "nitro")]
+use api_model::nitro::NitroEnclavesConversionRequest as PlatformConversionRequest;
+#[cfg(platform = "nitro")]
+use api_model::nitro::NitroEnclavesConversionResponse as PlatformConversionResponse;
```
sardok: Nit; you may tidy here if you want to with something like;
```
#[cfg(platform = "nitro")]
mod nitro_imports {
    pub(super) use crate::image_builder::nitro::enclave::EnclaveImageBuilder as PlatformEnclaveImageBuilder;
    pub(super) use crate::image_builder::nitro::parent::ParentImageBuilder as PlatformParentImageBuilder;
    pub(super) use crate::nitro::create_response;
    pub(super) use api_model::nitro::NitroEnclavesConversionRequest as PlatformConversionRequest;
    pub(super) use api_model::nitro::NitroEnclavesConversionResponse as PlatformConversionResponse;
}

#[cfg(platform = "nitro")]
use nitro_imports::*;
```

Or with `cfg-if` crate.
```
cfg_if::cfg_if! {
    if #[cfg(platform = "nitro")] {
        use crate::image_builder::nitro::enclave::EnclaveImageBuilder as PlatformEnclaveImageBuilder;
        use crate::image_builder::nitro::parent::ParentImageBuilder as PlatformParentImageBuilder;
        use crate::nitro::create_response;
        use api_model::nitro::NitroEnclavesConversionRequest as PlatformConversionRequest;
        use api_model::nitro::NitroEnclavesConversionResponse as PlatformConversionResponse;
   } else if #[cfg(platform = "snp")] {
        .....
   }
}
```

.github/workflows/build-solution.yml
```diff
@@ -57,13 +57,31 @@ jobs:
...
+        run: |
+          ./build-converter.sh
+
+      - name: Run Nitro unit tests
+        env:
+          FORTANIX_API_KEY: ${{ secrets.FORTANIX_API_KEY }}
+          OVERLAYFS_UNIT_TEST_API_KEY: ${{ secrets.OVERLAYFS_UNIT_TEST_API_KEY }}
+          SALMIAC_PLATFORM: nitro
+        run: |
+          ./unit-test-solution.sh
```
tvsfx:
```sh
  unit_test_dirs=(
      "vsock-proxy/enclave"
      "vsock-proxy/parent"
      "tools/container-converter"
      "tools/enclaveos-encrypted-fs"
  )
```
Do note that this will execute tests in `tools/enclaveos-encrypted-fs` and any non-platform-specific crate that we might add in the future multiple times.

I just took a look at `build-converter.sh` and I don't think it does anything useful; it was initially intended to compile `enclave_startup_bin_folder` but that is now taken care of by the `build.rs` script of `container-converter`.
Running `cargo build` in `tools/container-converter` doesn't test anything that the `cargo test` command in `unit-test-solution.sh` doesnt also do.
Thus , the simpler solution IMO is to follow Mihir's approach and remove the `build-converter.sh` file and its invocation entirely. 
We might want to run the `api-model` tests N times in that case.

Also, you could have used composite actions in your current solution:
 https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action#creating-an-action-metadata-file

tools/container-converter/src/lib.rs
rui-alm: a67de69f6fbfcca0a5011a1cb5022e15db5e9ccc

api-model/src/snp.rs
tvsfx: I think you missed this comment in the commit you reference 😆

rui-alm: Takes effort to miss that one :facepalm: 2a6000e567f5bc1f82bff4cea5b18418b41dfecb

tools/container-converter/src/lib.rs
tvsfx: Yeah, I like this better :) 

```rust
    pub(super) use crate::image_builder::enclave::nitro::EnclaveImageBuilder as PlatformEnclaveImageBuilder;
    pub(super) use crate::image_builder::parent::nitro::ParentImageBuilder as PlatformParentImageBuilder;
```
I would consider just putting these conditional imports here in `./image_builder/<enclave|parent>/mod.rs` respectively (similar for `api_model`); that's the most natural abstraction boundary to me.

rui-alm: 4adbebd82717b7be692d272a19ed23cd860392bb

tools/container-converter/src/image_builder/parent/nitro.rs
sardok: @rui-alm sorry missed this message. Yes i am fine.

tools/container-converter/src/image_builder/enclave/nitro.rs
sardok: Thats fine. The `create_manifest_file` i guess was expecting `self` when i left this comment. If `self` is not used in that function, yes it should be like this.

sardok (dismissed):

.github/workflows/build-solution.yml
rui-alm: 2c15a49e3a6877bf3a5e24ad31f768d3f58116a3, 8cd4050062ca62a45c5fb1bce121415d2458b47a

tvsfx (approved):

tools/container-converter/src/image_builder/parent/nitro.rs
rui-alm: @sardok Feel free to adjust this when you refactor the image builder.

PR #65, opened 2026-05-07
[RTE-940] Split Image tags and update build scripts

pavankalyan0424:
This PR includes following changes:
1. Restructured docker directory - We will now have docker/platform
2. Did split tags - Now images will be parent-base-platform, enclave-base-platform
3. Removed tools/ci-scripts dir - These scripts are not related to this repo.
4. Removed docker/build.sh, docker/build-enclave-kernel.sh as these are for internal purpose(Also incomplete/ not related to this repo).


docker/nitro/enclave-base/Dockerfile
anubhavg64, line 1: Earlier, we discussed using this Dockerfile for both AWS Nitro Enclaves and SEV-SNP confidential VMs without GPU passthrough. However, after the refactoring, I’m wondering whether we should include this in the snp directory as well.

pavankalyan0424: Let me check this again. May be I will move enclave-base one step behind.

pavankalyan0424: I missed this change, when I rebased the branch. Pushed a commit to address this.

docker/snp/build-conv-container.sh
```diff
@@ -0,0 +1,21 @@
+#TODO: RTE-941
```
rui-alm: RTE-941 is for the implementation of SNP image builders. I'd prefer for the base images and converter image to be dealt with as part of this PR.

At the moment this script won't work because there is `Dockerfile` to build the SNP converter, so let's add it and make sure we can produce all SNP images.

README.md
```diff
@@ -95,7 +95,7 @@ This guide allows you to build salmiac from source and convert your docker appli
 7. Make your application Nitro VM-capable by running container converter with the file from previous step.
    The converter by default pulls the input image and pushes the output image to remote repositories. These images are then cleaned up from the local docker cache. In our example, the output image push is disabled in the request json and to preserve the images in the docker cache, 'PRESERVE_IMAGES' environment variable is specified.
    ```bash
-      docker run --rm --name converter --user 0 --privileged -v /var/run/docker.sock:/var/run/docker.sock -e PRESERVE_IMAGES=input,result -v /tmp/req-files:/app converter --request-file /app/req.json
+      docker run --rm -e PARENT_IMAGE=parent-base-nitro -e ENCLAVE_IMAGE=enclave-base --name nitro-converter --user 0 --privileged -v /var/run/docker.sock:/var/run/docker.sock -e PRESERVE_IMAGES=input,result -v /tmp/req-files:/app nitro-converter --request-file /app/req.json
```
rui-alm: I think `-v /tmp/req-files:/app` will mount on `/app`, where the converter executable is, preventing it from being accessed. Can you confirm and if so adjust the mapping of the request files directory mount?

`ENCLAVE_IMAGE` is pointing at a generic `enclave-base` tag, but https://github.com/fortanix/salmiac/pull/62 has introduced the SNP enclave base image.

Can we have examples for both platforms, with their corresponding base image tags?

rui-alm (changes requested):

README.md
pavankalyan0424: > ENCLAVE_IMAGE is pointing at a generic enclave-base tag, but https://github.com/fortanix/salmiac/pull/62 has introduced the SNP enclave base image.

Reg. this, according to comment [here](https://github.com/fortanix/salmiac/blob/89883a6e0ba4c8f547954f7b693202a68ca4d73b/docker/enclave-base/Dockerfile#L5), I updated the PR.

I will look into adding examples for both platforms and fixing the docker run command.

rui-alm: Yes, that's correct for the non-GPU version. We also have https://github.com/fortanix/salmiac/blob/master/docker/enclave-base-snp-gpu/Dockerfile which is necessary for GPU workloads.

Our image tagging and publishing should reflect that and we can have examples on the readme for both GPU and non GPU deployments.

docker/snp/build-conv-container.sh
rui-alm: Is the SNP converter dockerfile missing from a6e71063eec6a5c2224fa9cc3e35499f2d325692?

README.md
```diff
@@ -16,7 +16,7 @@ Salmiac enhances Nitro Enclaves by enabling networking for external communicatio
 
 ## Quick Start Guide
 
-This guide allows you to build salmiac from source and convert your docker application into a one that can run in a nitro enclave. 
+This guide allows you to build salmiac from source and convert your docker application into a one that can run in a nitro enclave or snp enclave.
```
aditijannu: This README should have instructions only for building and running nitro enclaves for now. There are  parts of the snp enclave support that are not open sourced yet - for example building the enclave kernel.

run-application-tests.sh
```diff
@@ -15,7 +15,7 @@ echo "AWS_CREDENTIALS=$(/usr/bin/base64 --wrap=0 < ~/.aws/credentials)" >> docke
 ECR_PASSWORD=$(aws ecr get-login-password --region us-west-1)
 echo "ECR_PASSWORD=$ECR_PASSWORD" >> docker-env
 
-PARENT_IMAGE=fortanix/parent-base:3.0.0
+PARENT_IMAGE=fortanix/parent-base-nitro:3.0.0
```
aditijannu, line 18: You will need to update the fortanix dockerhub repository where parent-base image is stored.

README.md
pavankalyan0424: Ok, I will update it back.

run-application-tests.sh
pavankalyan0424: Yes, i plan to push.

docker/snp/build-conv-container.sh
pavankalyan0424: yes I messed that commit. I appended and re-pushed.

README.md
pavankalyan0424: According to discussion [here](https://github.com/fortanix/salmiac/pull/65#discussion_r3226915639), we plan to not add snp documentation for now.

docker/snp/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+dpkg -x ovmf*.deb ovmf_dir && \
+cp ovmf_dir/usr/share/ovmf/OVMF.amdsev.fd /blob
+
+FROM ubuntu:24.04
+
+RUN mkdir -p /var/log/snp_enclaves
+
+WORKDIR /app
+
+COPY staging/server staging/parent-base.tar staging/enclave-base.tar staging/enclave-base-gpu.tar /app/
```
rui-alm, line 22: @sardok I believe the converter is accessing the images via the env vars `ENCLAVE_IMAGE` and `PARENT_IMAGE`, but please confirm when you have a chance. If so, have there been discussions recently about changing this behaviour?

@aditijannu Do we mind packing these images to potentially be used in the future, or should we remove them for now to keep the converter image smaller?

README.md
rui-alm: _instance_

docker/snp/Dockerfile
aditijannu: We will have to re-introduce the tar files soon if we get rid of it now, we want the converter to have the ability to convert images without external  network access. So if the ENCLAVE_IMAGE and PARENT_IMAGE can't be downloaded from docker hub, converter is expected to use these tar files as the fallback option. This feature may not be working as expected at the moment and needs to be fixed.
So maybe keep it since we want to use it in the future?

rui-alm: Sounds good!

rui-alm (approved):

PR #66, opened 2026-05-07
Rte 916/sev snp measure

anubhavg64:
What's in this PR:

- tools/sev-snp-measure/ - small Python wrapper around the sev-snp-measure tool.
- tools/container-converter/.../snp_measurement.rs - Rust function                        compute_snp_launch_measurement that runs the Python wrapper and returns the result to the caller.
- docker/snp/Dockerfile - minimal Dockerfile that installs the tooling into the SNP converter image. RTE-941 extends this with everything else.
Depends on #72

### DIFF
--- a/docker/snp/Dockerfile
+++ b/docker/snp/Dockerfile
@@ -12,6 +12,16 @@ RUN apt-get download ovmf && \
     dpkg -x ovmf*.deb ovmf_dir && \
     cp ovmf_dir/usr/share/ovmf/OVMF.amdsev.fd /blob
 
+FROM debian:bookworm-slim AS sev_snp_measurement
+RUN apt-get update && apt-get install -y \
+        python3 \
+        python3-pip \
+        git \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY staging/sev-snp-measure/requirements.txt /tmp/requirements.txt
+RUN pip install --break-system-packages --target /opt/sev-snp-dist -r /tmp/requirements.txt
+
 FROM ubuntu:24.04
 
 ENV DEBIAN_FRONTEND=noninteractive
@@ -19,13 +29,13 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && \
     apt-get install -y \
         cryptsetup-bin \
-        python3 \
-        python3-pip \
-        git
+        python3
+
+COPY --from=sev_snp_measurement /opt/sev-snp-dist /opt/sev-snp-dist
+ENV PYTHONPATH="/opt/sev-snp-dist"
 
 COPY staging/sev-snp-measure /opt/sev-snp-measure
-RUN pip install --break-system-packages -r /opt/sev-snp-measure/requirements.txt && \
-    chmod +x /opt/sev-snp-measure/measure.py
+RUN chmod +x /opt/sev-snp-measure/measure.py
 
 RUN mkdir -p /var/log/enclave-os
 
--- a/docker/snp/Dockerfile
+++ b/docker/snp/Dockerfile
@@ -5,20 +5,27 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update
 RUN mkdir -p /blob
 
 RUN apt-get download systemd-boot-efi && \
-dpkg -x systemd-boot-efi*.deb system_boot_efi_dir &&\
-cp system_boot_efi_dir/usr/lib/systemd/boot/efi/linuxx64.efi.stub /blob
-
+    dpkg -x systemd-boot-efi*.deb system_boot_efi_dir && \
+    cp system_boot_efi_dir/usr/lib/systemd/boot/efi/linuxx64.efi.stub /blob
 
 RUN apt-get download ovmf && \
-dpkg -x ovmf*.deb ovmf_dir && \
-cp ovmf_dir/usr/share/ovmf/OVMF.amdsev.fd /blob
+    dpkg -x ovmf*.deb ovmf_dir && \
+    cp ovmf_dir/usr/share/ovmf/OVMF.amdsev.fd /blob
 
 FROM ubuntu:24.04
 
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y cryptsetup-bin
+    apt-get install -y \
+        cryptsetup-bin \
+        python3 \
+        python3-pip \
+        git
+
+COPY staging/sev-snp-measure /opt/sev-snp-measure
+RUN pip install --break-system-packages -r /opt/sev-snp-measure/requirements.txt && \
+    chmod +x /opt/sev-snp-measure/measure.py
 
 RUN mkdir -p /var/log/enclave-os
 
--- a/tools/container-converter/src/image_builder/enclave/snp.rs
+++ b/tools/container-converter/src/image_builder/enclave/snp.rs
@@ -114,11 +114,7 @@ fn create_initramfs(
     enclave_settings: &EnclaveSettings,
     enclave_manifest_data: &[u8],
 ) -> std::result::Result<(), IoError> {
-    let run_cmd = GenericEnclaveImageBuilder::enclave_command_string(
-        enclave_settings,
-        &Path::new(INSTALLATION_DIR),
-        "enclave",
-    );
+    let run_cmd = get_run_cmd(&enclave_settings);
 
     let mut blobs_dir = PathBuf::from(format!("{}/blobs", INSTALLATION_DIR));
     let init = {
@@ -136,7 +132,7 @@ fn create_initramfs(
     };
 
     let mut fs_tree = FsTree::new()
-        .add_file("env", Cursor::new(enclave_settings.env_vars.join("\n")))
+        .add_file("env", Cursor::new(get_env_vars(&enclave_settings.env_vars)))
         .add_file("cmd", Cursor::new(run_cmd))
         .add_executable("init", Cursor::new(init));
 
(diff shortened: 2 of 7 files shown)


tools/sev-snp-measure/measure.py
```diff
@@ -0,0 +1,57 @@
...
+
+import argparse
+import sys
+from pathlib import Path
+
+from sevsnpmeasure.guest import calc_launch_digest
+from sevsnpmeasure.sev_mode import SevMode
+
+FORTANIX_VME_VCPU_SIG = 0
+FORTANIX_VME_GUEST_FEATURES = 1
```
rui-alm, line 18: Has there been an analysis to the guest features? https://github.com/virtee/sev-snp-measure#sev-snp-guest-feature-field-values

anubhavg64: Good catch, The QEMU launcher in RTE-917 (vsock-proxy/parent/src/platform/snp.rs) currently doesn't set guest-features= explicitly: 

So QEMU 10.2.2's default applies at runtime. Our measure.py hardcodes guest_features=1 (mirrored from the existing fortanixvme-snp-sign script in roche), but that's an assumption, not a verified match.                                                      
How would you like us to proceed:
1. Remove guest_features from measure.py entirely and let sev-snp-measure's default apply, and accept the same default on both sides.                                                 
2. Keep guest_features=1 explicit and we'd also set guest-features=1 in the QEMU launcher (RTE-917) so both sides are deterministic.

anubhavg64: @MihirLuthra I have used `0x21` which is the default, and added a TODO to revisit this in next phase.

tools/container-converter/src/image_builder/enclave/snp.rs
sardok: Let's rename this as `initramfs_file_path`.

anubhavg64: Done! 5f41dd5

tools/container-converter/src/image_builder/enclave/snp_measurement.rs
```diff
@@ -0,0 +1,62 @@
+/* Copyright (c) Fortanix, Inc.
```
sardok, line 1: All python script usage/wrapper code could be replaced by `sev` crate: See https://github.com/virtee/sev/blob/76535be354648a3e4adf00631fdf4941ab51fb88/src/measurement/snp.rs#L232.
But i guess we want to have our modification about signing on `sev-snp-measurement` tool in future, thats why i guess we preferred py tool over rust crate?

anubhavg64: Yeah, pretty much, kept the Python tool since the fortanix fork has the signing bits we might want later on.

tools/sev-snp-measure/measure.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
```
sardok, line 1: I don't understand why we would need this. This file is pretty much covered by `sev-snp-measure` bin?

docker/snp/Dockerfile
```diff
@@ -5,20 +5,27 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update
...
 
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y cryptsetup-bin
+    apt-get install -y \
+        cryptsetup-bin \
+        python3 \
+        python3-pip \
+        git
```
sardok: IMO, it is better to build the `sev-snp-measure` in an intermediate context and copy the artifacts to be more cache friendly and less dependency in final image (like git, python3-pip);

```
FROM ubuntu:25.10 AS apt_extract
// Existing instructions


FROM debian:bookworm-slim AS sev_snp_measurement

RUN apt-get update && apt-get install -y \
    python3 \
    python3-pip \
    git \
    curl \
    && rm -rf /var/lib/apt/lists/*

COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/

RUN uv pip install --system --target /opt/sev-snp-dist git+https://github.com/virtee/sev-snp-measure.git

FROM ubuntu:24.04

ENV DEBIAN_FRONTEND=noninteractive

RUN apt-get update && \
    apt-get install -y \
        cryptsetup-bin \
        python3

COPY --from=sev_snp_measurement /opt/sev-snp-dist /opt/sev-snp-dist

ENV PYTHONPATH="/opt/sev-snp-dist"
ENV PATH="/opt/sev-snp-dist/bin:$PATH"

// The rest follows the same
```

tools/container-converter/src/image_builder/enclave/snp_measurement.rs
sardok: Ok but can't we do the signing modifications within enclave image builder with the necessary crypto libs?

tools/sev-snp-measure/measure.py
anubhavg64: Well I somewhat agree. Do you suggest to drop measure.py and call the upstream sev-snp-measure CLI directly from the Rust side, passing --vcpu-sig 0 --guest-features 0x21 as flags?
That lets me delete the whole tools/sev-snp-measure/ directory and reduce the Dockerfile change to a single pip install.

sardok: I've just noticed that we removed `sev-snp-measure` cli from our fork. Ok, let's keep it as is for now.

docker/snp/Dockerfile
anubhavg64: Done! e76bbef

tools/sev-snp-measure/measure.py
```diff
@@ -0,0 +1,58 @@
...
+import argparse
+import sys
+from pathlib import Path
+
+from sevsnpmeasure.guest import calc_launch_digest
+from sevsnpmeasure.sev_mode import SevMode
+
+FORTANIX_VME_VCPU_SIG = 0
+# TODO: needs security hardening.
+FORTANIX_VME_GUEST_FEATURES = 0x21
```
tvsfx: Why is this different from https://github.com/fortanix/sev-snp-measure/blob/887f3812b61943672f6d5420043192195b2584ef/fortanixvme-snp-sign#L79 ?

tools/container-converter/src/image_builder/enclave/snp_measurement.rs
tvsfx: > Yeah, pretty much, kept the Python tool since the fortanix fork has the signing bits we might want later on.

If we want to use those, wouldn't it make more sense to file the changes required for `measure.py` as a PR to extend argument parsing on https://github.com/fortanix/sev-snp-measure?

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -90,9 +92,24 @@ impl<'a> EnclaveImageBuilder<'a> {
...
+                BLOBS_SUBDIR_GPU_DISABLED
+            })
+            .join("bzImage");
+
+        compute_snp_launch_measurement(&SnpMeasurementInputs {
+            ovmf: &ovmf_path,
+            kernel: &kernel_path,
+            initrd: Some(&initramfs_file_path),
+            cmdline: None,
+            vcpus: 1,
```
tvsfx: https://github.com/fortanix/salmiac/pull/67/changes#diff-9324b41922d8b9a87600774d49a142f04b95ee2bad7460cb17e5b895d355a6daR12
It seems like Mihir is picking 2 as a hardcoded value; we need to discuss this

tools/container-converter/src/image_builder/enclave/snp_measurement.rs
```diff
@@ -0,0 +1,62 @@
...
+) -> Result<SNPEnclavesMeasurements> {
+    let mut args: Vec<OsString> = vec![
+        "--ovmf".into(),
+        inputs.ovmf.as_os_str().to_os_string(),
+        "--kernel".into(),
+        inputs.kernel.as_os_str().to_os_string(),
+        "--vcpus".into(),
+        inputs.vcpus.to_string().into(),
+    ];
+    if let Some(initrd) = inputs.initrd {
```
sardok, line 36: Once we figure out how to get `cmdline`, these variables `initrd` and `cmdline` should not be optional otherwise if passed empty, the measurements check would fail.

tools/sev-snp-measure/measure.py
anubhavg64: From Kernel 6.6 the default value is 0x21.
https://github.com/fortanix/salmiac/pull/66#discussion_r3273471618

tools/container-converter/src/image_builder/enclave/snp.rs
anubhavg64: Done! 81ef489c858d7c43a1ad28642af3b17ae804589c

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -90,9 +92,24 @@ impl<'a> EnclaveImageBuilder<'a> {
...
+            .join("bzImage");
+
+        compute_snp_launch_measurement(&SnpMeasurementInputs {
+            ovmf: &ovmf_path,
+            kernel: &kernel_path,
+            initrd: Some(&initramfs_file_path),
+            cmdline: None,
+            vcpus: 2,
         })
+        .await
```
sardok: Its better to move these lines in a new function something like `compute_launch_measurements`.

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -90,9 +92,24 @@ impl<'a> EnclaveImageBuilder<'a> {
             kind: ConverterErrorKind::EnclaveImageCreation,
         })?;
 
-        Ok(SNPEnclavesMeasurements {
-            launch_measurement: HexString::new(vec![0; 48]),
+        let blobs_dir = Path::new(INSTALLATION_DIR).join("blobs");
+        let ovmf_path = blobs_dir.join("OVMF.amdsev.fd");
```
sardok: Better to assign `OVMF.amdsev.fd` to a const string like `EnclaveImageBuilder::INIT_BIN`.

tools/container-converter/src/image_builder/enclave/snp.rs
anubhavg64: Done! 610ef054053af1284b1b8738c5015f4a273adff7

tools/container-converter/src/image_builder/enclave/snp.rs
anubhavg64: Done! c48c09f3e08e5ab1b4d232881dc2a28bb9845e47

pavankalyan0424 (dismissed):

PR #67, opened 2026-05-08
RTE-917: Add support for platforms SNP and Simulator in vsock-proxy

MihirLuthra:
Reference changes for supporting SNP in vsock-proxy. These are incomplete right now as we need to have a consensus on things converter expects.
For testing the current version, I used some scripts to generate a temporary setup. The setup is basically a test parent image based on `parent-base-snp` that launches vsock-proxy parent, which in turn starts qemu. The qemu command is given an initramfs (based on `enclave-base-snp`). On startup of vm, it runs a vsock-proxy enclave which is given an `enclave-settings.json` pointing to a hello world app (the app is in the blockfile).
Note: I made some changes to qemu args (like policy and FMS values) after my testing. (I will update description once I have tested latest)

This is currently not ready for review - it needs some cleanup and I am not done confirming on the exact parameters used in qemu command. But feel free to add comments if any.

### DIFF
--- a/vsock-proxy/parent/src/parent.rs
+++ b/vsock-proxy/parent/src/parent.rs
@@ -22,7 +22,7 @@ use shared::socket::{AsyncReadLvStream, AsyncWriteLvStream};
 use shared::tap::{start_tap_loops, PRIVATE_TAP_MTU, PRIVATE_TAP_NAME};
 use shared::{
     cleanup_tokio_tasks, run_subprocess, run_subprocess_with_output_setup, with_background_tasks, AppLogPortInfo,
-    CommandOutputConfig, StreamType, DNS_RESOLV_FILE, HOSTNAME_FILE, HOSTS_FILE, NS_SWITCH_FILE, VSOCK_PARENT_CID,
+    CommandOutputConfig, StreamType, DNS_RESOLV_FILE, HOSTNAME_FILE, HOSTS_FILE, NS_SWITCH_FILE, VSOCK_LISTENER_CID,
 };
 use tokio::net::TcpListener;
 use tokio::task::JoinHandle;
@@ -607,8 +607,8 @@ async fn create_vsock_stream(port: u32) -> Result<AsyncVsockStream, String> {
 }
 
 pub(crate) fn listen_to_parent(port: u32) -> Result<AsyncVsockListener, String> {
-    AsyncVsockListener::bind(VSOCK_PARENT_CID, port)
-        .map_err(|_| format!("Could not bind to cid: {}, port: {}", VSOCK_PARENT_CID, port))
+    AsyncVsockListener::bind(VSOCK_LISTENER_CID, port)
+        .map_err(|_| format!("Could not bind to cid: {}, port: {}", VSOCK_LISTENER_CID, port))
 }
 
 pub(crate) async fn accept(listener: &mut AsyncVsockListener) -> Result<AsyncVsockStream, String> {
--- a/vsock-proxy/parent/src/parent.rs
+++ b/vsock-proxy/parent/src/parent.rs
@@ -7,7 +7,7 @@
 use std::collections::HashMap;
 use std::fs::{File, OpenOptions};
 use std::io::{BufRead, BufReader, Write};
-use std::net::{IpAddr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::{env, fs};
@@ -26,6 +26,7 @@ use shared::{
 };
 use tokio::net::TcpListener;
 use tokio::task::JoinHandle;
+use tokio::time::{sleep, Duration, Instant};
 use tokio_vsock::{VsockListener as AsyncVsockListener, VsockStream as AsyncVsockStream};
 
 use crate::network::{
@@ -55,6 +56,10 @@ const VSOCK_PARENT_PORT: u32 = 5006;
 const NAMESERVER_KEYWORD: &'static str = "nameserver";
 
 const CLIENT_LOG_STREAMS: [StreamType; 2] = [StreamType::Stdout, StreamType::Stderr];
+
+const NBD_SERVER_READY_TIMEOUT: Duration = Duration::from_secs(5);
+const NBD_SERVER_READY_RETRY_INTERVAL: Duration = Duration::from_millis(50);
+const TCP_LISTEN_STATE: &str = "0A";
 
 async fn message_handler(enclave: &mut AsyncVsockStream) -> Result<UserProgramExitStatus, String> {
     loop {
@@ -98,7 +103,7 @@ pub(crate) async fn run(args: ParentConsoleArguments) -> Result<UserProgramExitS
     let tap_l3_address = setup_result.private_tap.tap_l3_address.ip();
 
     let mut log_listeners = setup_log_listeners(tap_l3_address).await?;
-    let mut background_tasks = start_background_tasks(setup_result)?;
+    let mut background_tasks = start_background_tasks(setup_result).await?;
 
     let log_ports = get_log_sock_addrs(&mut log_listeners)?;
     info!("Client log listeners set up.");
@@ -326,6 +331,39 @@ async fn run_nbd_server(port: u16) -> Result<(), String> {
     }
 }
 
+async fn wait_for_nbd_server(address: IpAddr, port: u16) -> Result<(), String> {
+    let address = match address {
+        IpAddr::V4(address) => address,
+        IpAddr::V6(address) => return Err(format!("IPv6 NBD readiness check is not supported for {address}")),
+    };
+
+    let start = Instant::now();
+
+    while start.elapsed() < NBD_SERVER_READY_TIMEOUT {
+        if tcp_port_is_listening(address, port)? {
+            return Ok(());
+        }
+
+        sleep(NBD_SERVER_READY_RETRY_INTERVAL).await;
+    }
+
+    Err(format!(
+        "nbd-server did not start listening on {address}:{port} within {:?}",
+        NBD_SERVER_READY_TIMEOUT
+    ))
+}
+
+fn tcp_port_is_listening(address: Ipv4Addr, port: u16) -> Result<bool, String> {
+    let expected = format!("{:08X}:{port:04X}", u32::from_le_bytes(address.octets()));
+    let proc_net_tcp = fs::read_to_string("/proc/net/tcp").map_err(|err| format!("failed to read /proc/net/tcp: {err}"))?;
+
+    Ok(proc_net_tcp.lines().skip(1).any(|line| {
+        let fields = line.split_whitespace().collect::<Vec<_>>();
+
+        fields.len() > 3 && fields[1].eq_ignore_ascii_case(&expected) && fields[3] == TCP_LISTEN_STATE
+    }))
+}
+
 /// Start the dnsmasq process. dnsmasq is a DNS server/proxy. We configure dnsmasq to listen
 /// on the fortanix-tap0 device, so that device must be set up first before we start dnsmasq.
 /// dnsmasq will refuse to run if the configured interface is not present when it starts.
@@ -372,7 +410,7 @@ async fn run_log_listeners(listeners_info: Vec<(TcpListener, StreamType)>) -> Re
     Ok(tasks)
 }
 
-fn start_background_tasks(
+async fn start_background_tasks(
     parent_setup_result: ParentSetupResult,
 ) -> Result<FuturesUnordered<JoinHandle<Result<(), String>>>, String> {
     let result = FuturesUnordered::new();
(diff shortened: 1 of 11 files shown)


vsock-proxy/parent/src/platform/snp.rs
```diff
@@ -0,0 +1,130 @@
...
+const DEFAULT_CPU: &str = "host,-tsa-sq-no,-tsa-l1-no,family=0,model=0,stepping=0";
+
+// https://docs.amd.com/v/u/en-US/58207-using-sev-with-amd-epyc-processors
+//
+// "Since SNP is only supported from processor series 7003 and newer, the c-bit (cbitpos) will always be 51"
+const DEFAULT_CBITPOS: &str = "51";
+const DEFAULT_REDUCED_PHYS_BITS: &str = "1";
+
+// We want SMT disabled
+const DEFAULT_POLICY: &str = "0x20000";
```
rui-alm, line 29: The SNP policy is `0x30000`, isn't it?

vsock-proxy/parent/src/platform/snp.rs
```diff
@@ -0,0 +1,130 @@
...
+
+const CPU_COUNT: &str = "2";
+const MEMORY_SIZE: &str = "8192M";
+
+/// - Use host passthrough, see https://www.qemu.org/docs/master/system/i386/cpu.html
+/// - Unset TSA_L1_NO and TSA_SQ_NO:
+///   See: https://www.amd.com/content/dam/amd/en/documents/resources/bulletin/technical-guidance-for-mitigating-transient-scheduler-attacks.pdf
+///   "the SEV-SNP FW will not allow the hypervisor to set the TSA_L1_NO or TSA_SQ_NO CPUID bits"
+/// - TODO(before merge): Confirm family=0,model=0,stepping=0
+const DEFAULT_CPU: &str = "host,-tsa-sq-no,-tsa-l1-no,family=0,model=0,stepping=0";
```
rui-alm: The `-tsa-sq-no,-tsa-l1-no` options, which QEMU versions support this? Do they need to be explicitly set or does the SNP firmware deal with it for all affected CPUs?

vsock-proxy/parent/src/platform/snp.rs
```diff
@@ -0,0 +1,130 @@
...
+// https://docs.amd.com/v/u/en-US/58207-using-sev-with-amd-epyc-processors
+//
+// "Since SNP is only supported from processor series 7003 and newer, the c-bit (cbitpos) will always be 51"
+const DEFAULT_CBITPOS: &str = "51";
+const DEFAULT_REDUCED_PHYS_BITS: &str = "1";
+
+// We want SMT disabled
+const DEFAULT_POLICY: &str = "0x20000";
+
+// TODO(before merge): Make UKI instead, and need to check on final kernel cmdline
```
rui-alm: What do we mean by UKI? What exactly are we looking at unifying?

vsock-proxy/parent/src/platform/snp.rs
```diff
@@ -0,0 +1,130 @@
...
+const KERNEL_CMDLINE: &str = "console=ttyS0 rdinit=/init loglevel=7";
+const KERNEL_PATH: &str = "/opt/fortanix/enclave-os/kernel";
+const INITRAMFS_PATH: &str = "/opt/fortanix/enclave-os/initramfs.cpio.gz";
```
rui-alm: Is there a requirement for the initramfs to be compressed? Or is this just because you happen to be testing with a compressed initramfs at the moment?

rui-alm (commented):
I know this is still a draft, but there's a few topics I'd like to clarify.

vsock-proxy/parent/src/platform/snp.rs
MihirLuthra: Looking at the date of the commit, I think this should be supported 10.2 onwards: https://github.com/qemu/qemu/commit/c79a35acadee784610aed40134a12738381b4fba. In our parent-base-snp, we have `QEMU_VERSION=v10.2.2` so works okay there.

If we don't set `-tsa-sq-no,-tsa-l1-no` and use only `host`, firmware will complain about the same:
```
qemu-system-x86_64: SNP_LAUNCH_UPDATE ret=-5 fw_error=22 'Invalid parameter'
qemu-system-x86_64: SEV-SNP: CPUID validation failed for function 0x80000021, index: 0x0, provided: eax:0x18000167, ebx: 0x00000000, ecx: 0x00000006, edx: 0x00000000, expected: eax:0x18000167, ebx: 0x00000000, ecx: 0x00000000, edx: 0x00000000
qemu-system-x86_64: SEV-SNP: failed update CPUID page
```

vsock-proxy/parent/src/platform/snp.rs
MihirLuthra: `0x20000` would mean that we disable SMT as well. There is a thread between Jethro/Joep about mitigation of some CVEs wrt SMT in AMD processors, and concluding that we should disable SMT.

vsock-proxy/parent/src/platform/snp.rs
MihirLuthra: I thought of staying consistent with EDP, but I don't see a good reason. I don't think we want to sign anything here, so whether we make a UKI or let artifacts be separate depends on how we prefer it in our code.

vsock-proxy/parent/src/platform/snp.rs
MihirLuthra: Would prevent bigger file size/easier to copy around. No major reason.

vsock-proxy/parent/src/platform/snp.rs
jethrogb: In my current plans for IGVM we wouldn't use EFI and therefore won't use UKI.

vsock-proxy/parent/src/platform/snp.rs
```diff
@@ -0,0 +1,186 @@
...
+const SERIAL: &str = "telnet:0.0.0.0:4321,server,wait";
+
+// TODO: We need to see how we will assign guest CID when we support multiple
+// containers on single host
+const VSOCK_DEVICE: &str = "vhost-vsock-pci,guest-cid=3";
+
+const IOMMUFD_ID: &str = "iommufd0";
+const IOMMUFD_OBJECT: &str = "iommufd,id=iommufd0";
+const GPU_ROOT_PORT: &str = "pcie-root-port,id=pci.1,bus=pcie.0";
+const FW_CFG_MMIO64: &str = "name=opt/ovmf/X-PciMmio64Mb,string=262144";
```
MihirLuthra, line 48: GPU enabling changes are temporary, and testing is in progress. We are encountering some issues which are being debugged

MihirLuthra, line 48: Also, the inclination is to passthrough all GPUs by default for phase 1. This PR doesn't reflect that yet.

vsock-proxy/parent/src/parent.rs
```diff
@@ -326,6 +331,39 @@ async fn run_nbd_server(port: u16) -> Result<(), String> {
...
+        sleep(NBD_SERVER_READY_RETRY_INTERVAL).await;
+    }
+
+    Err(format!(
+        "nbd-server did not start listening on {address}:{port} within {:?}",
+        NBD_SERVER_READY_TIMEOUT
+    ))
+}
+
+fn tcp_port_is_listening(address: Ipv4Addr, port: u16) -> Result<bool, String> {
```
MihirLuthra, line 356: This is not foolproof. To prevent nbd-server race condition, we are just seeing if something is listening on address:port.

vsock-proxy/parent/src/platform/simulator.rs
```diff
@@ -0,0 +1,82 @@
+/* Copyright (c) Fortanix, Inc.
```
MihirLuthra, line 1: I will remove simulator file for this PR. Can add later if useful

vsock-proxy/parent/src/parent.rs
jethrogb: This seems unrelated to SNP?

MihirLuthra: This is a fix for RTE-952. This is not directly related to SNP support but this race condition shows up almost everytime in case of SNP.

vsock-proxy/shared/src/lib.rs
```diff
@@ -79,9 +80,21 @@ pub fn vec_to_ip6(vec: &[u16]) -> Result<Ipv6Addr, String> {
...
+#[cfg(any(platform = "simulator", platform = "snp"))]
+pub const VSOCK_PARENT_CID: u32 = 2;
+
+#[cfg(platform = "nitro")]
+// Context identifier of the parent (https://docs.aws.amazon.com/enclaves/latest/user/nitro-enclave-concepts.html)
+pub const VSOCK_LISTENER_CID: u32 = VSOCK_PARENT_CID;
+
+#[cfg(any(platform = "simulator", platform = "snp"))]
+// TODO(before merge): don't hardcode u32::MAX
+pub const VSOCK_LISTENER_CID: u32 = u32::MAX; // VMADDR_CID_ANY
```
sardok: `VMADDR_CID_ANY` is available in `tokio_vsock`.

vsock-proxy/parent/src/parent.rs
```diff
@@ -326,6 +331,39 @@ async fn run_nbd_server(port: u16) -> Result<(), String> {
...
+    let expected = format!("{:08X}:{port:04X}", u32::from_le_bytes(address.octets()));
+    let proc_net_tcp = fs::read_to_string("/proc/net/tcp").map_err(|err| format!("failed to read /proc/net/tcp: {err}"))?;
+
+    Ok(proc_net_tcp.lines().skip(1).any(|line| {
+        let fields = line.split_whitespace().collect::<Vec<_>>();
+
+        fields.len() > 3 && fields[1].eq_ignore_ascii_case(&expected) && fields[3] == TCP_LISTEN_STATE
```
jethrogb:
```suggestion
        let mut fields = line.split_whitespace();
        let Some(listen_addr) = fields.nth(1) else { return false };
        let Some(listen_state) = fields.nth(1) else { return false };

        listen_addr.eq_ignore_ascii_case(&expected) && listen_state == TCP_LISTEN_STATE
```

vsock-proxy/parent/src/parent.rs
jethrogb: I'm not a fan of this solution. I've left some comments on the ticket.

vsock-proxy/parent/src/platform/simulator.rs
```diff
@@ -0,0 +1,82 @@
...
+use std::path::Path;
+
+use shared::run_subprocess;
+
+use super::{GuestLaunchResult, GuestTasks};
+
+const QEMU_BINARY: &str = "qemu-system-x86_64";
+
+const KERNEL_PATH: &str = "/opt/fortanix/enclave-os/kernel";
+const INITRAMFS_PATH: &str = "/opt/fortanix/enclave-os/initramfs.cpio.gz";
```
sardok, line 15: This is `/opt/fortanix/enclave-os/initramfs.gz` in converted image.

vsock-proxy/parent/src/platform/simulator.rs
```diff
@@ -0,0 +1,82 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+use std::path::Path;
+
+use shared::run_subprocess;
+
+use super::{GuestLaunchResult, GuestTasks};
+
+const QEMU_BINARY: &str = "qemu-system-x86_64";
+
+const KERNEL_PATH: &str = "/opt/fortanix/enclave-os/kernel";
```
sardok, line 14: This is `/opt/fortanix/enclave-os/bzImage` in converted image.

vsock-proxy/parent/src/platform/simulator.rs
sardok: @MihirLuthra letme know if you don't like the filenames in converted image.

vsock-proxy/parent/src/parent.rs
MihirLuthra: Let me check and update

vsock-proxy/parent/src/platform/simulator.rs
MihirLuthra: I don't have a strong feeling as it is internal, but `cpio.gz` and `kernel` might be more readable.

vsock-proxy/parent/src/platform/snp.rs
```diff
@@ -0,0 +1,186 @@
...
+        CPU_COUNT,
+        "-m",
+        MEMORY_SIZE,
+        "-object",
+        &memory_backend,
+        "-object",
+        &snp_guest,
+        "-bios",
+        OVMF_PATH,
+        "-kernel",
```
tvsfx, line 122: Note that unlike for our rust-sgx repo (see [here](https://github.com/fortanix/rust-sgx/blob/1b3983c5e5bad113f75db1817be365786be4ef96/fortanix-vme/tools/elf2uki/src/main.rs?plain=1#L127-L128)), this QEMU invocation does not rely on an EFI boot stub. This means that our kernel will need to have `CONFIG_EFI_STUB=y` set, so that it boots without a bootloader/boot stub. The default ubuntu image does (hence why this works), but a future hardened image might not - though `y` is the default. Should we explicitly set this during kernel build?

tvsfx: So the reason we had to do this before is that we had a UKI image, with afaict requires a UEFI stub. Now we don't.

tvsfx: Ah, I forgot the other reason I mentioned this. That does mean we have no need for the efi-stub [here](https://github.com/fortanix/salmiac/blob/f02d858116b86e6bd0dffa9263283d485a03049c/docker/snp/Dockerfile?plain=1#L7-L9) @pavankalyan0424

PR #68, opened 2026-05-11
RTE-950: cargo fmt all Rust crates and check in CI

MihirLuthra:


tvsfx:
Looks good! Should we start a `.git-blame-ignore-revs` file in a separate commit to ignore the first commit, and then ff-merge both commits? An added benefit is that Github picks up on this automatically if we give it that name.

MihirLuthra:
> Looks good! Should we start a `.git-blame-ignore-revs` file in a separate commit to ignore the first commit, and then ff-merge both commits? An added benefit is that Github picks up on this automatically if we give it that name.

Good idea, I didn't know about this. I have added a `.git-blame-ignore-revs`.

tvsfx:
> Good idea, I didn't know about this. I have added a `.git-blame-ignore-revs`.

Ok, thanks! Now we just got to make sure to merge correctly; otherwise the commit hash we list is wrong :) Ideally the formatting changes are all in a single commit that ends up on master, so that whether or not we use `--first-parent` with git-blame we always ignore this commit. So either we just fast-forward merge as I said and make sure to put RTE-950 in each commit's text, or we split the PR in two and squash/ff-merge the formatting changes

tvsfx (dismissed):

aditijannu:
Discussed with @tvsfx - the default merge strategy for salmiac is to squash and merge. @MihirLuthra we merge only the first commit now and create another PR for the remaining two commits that add sanity.sh in CI and ignores the cargo fmt merged commit hash.

aditijannu (approved):

tvsfx (approved):

PR #72, opened 2026-05-15
[RTE-941] Add parent & enclave converter support for snp.

sardok:
Adds parent & enclave image converter implementation for SNP.

Enclave converter:
Based on the type request (gpu or no-gpu), it selects the correct enclave-base image. Creates an initramfs along with block file. The structure of `initramfs` is aligned with what `init` binary expects.

Parent converter:
Copies created artifacts by enclave converter as well as necessary blobs from `/opt/fortanix/enclave-os` to converted image.

### DIFF
--- a/docker/snp/Dockerfile
+++ b/docker/snp/Dockerfile
@@ -27,9 +27,14 @@ WORKDIR /app
 COPY staging/server staging/parent-base.tar staging/enclave-base.tar staging/enclave-base-gpu.tar /app/
 
 COPY --from=apt_extract /blob /opt/fortanix/enclave-os/blobs
-COPY init /opt/fortanix/enclave-os/blobs/init
+
+# Copy gpu disabled artifacts
+COPY kernel_disabled_gpu/init /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/init
 COPY kernel_disabled_gpu/.config /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/bzImage.config
 COPY kernel_disabled_gpu/bzImage /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/bzImage
+
+# Copy gpu enabled artifacts
+COPY kernel_enabled_gpu/init /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/init
 COPY kernel_enabled_gpu/.config /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/bzImage.config
 COPY kernel_enabled_gpu/bzImage /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/bzImage
 COPY kernel_enabled_gpu/nvidia-uvm.ko /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/nvidia-uvm.ko
--- a/docker/snp/Dockerfile
+++ b/docker/snp/Dockerfile
@@ -1,5 +1,5 @@
 # Download OVMF and EFI boot stubb
-FROM ubuntu:25.10 as apt_extract
+FROM ubuntu:25.10 AS apt_extract
 RUN DEBIAN_FRONTEND=noninteractive apt-get update
 
 RUN mkdir -p /blob
@@ -15,18 +15,24 @@ cp ovmf_dir/usr/share/ovmf/OVMF.amdsev.fd /blob
 
 FROM ubuntu:24.04
 
-RUN mkdir -p /var/log/snp_enclaves
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && \
+    apt-get install -y cryptsetup-bin
+
+RUN mkdir -p /var/log/enclave-os
 
 WORKDIR /app
 
 COPY staging/server staging/parent-base.tar staging/enclave-base.tar staging/enclave-base-gpu.tar /app/
 
-COPY --from=apt_extract /blob /usr/share/snp_enclaves/blobs
-COPY kernel_disabled_gpu/.config /usr/share/snp_enclaves/blobs/kernel_disabled_gpu/bzImage.config
-COPY kernel_disabled_gpu/bzImage /usr/share/snp_enclaves/blobs/kernel_disabled_gpu/bzImage
-COPY kernel_enabled_gpu/.config /usr/share/snp_enclaves/blobs/kernel_enabled_gpu/bzImage.config
-COPY kernel_enabled_gpu/bzImage /usr/share/snp_enclaves/blobs/kernel_enabled_gpu/bzImage
-COPY kernel_enabled_gpu/nvidia-uvm.ko /usr/share/snp_enclaves/blobs/kernel_enabled_gpu/nvidia-uvm.ko
-COPY kernel_enabled_gpu/nvidia.ko /usr/share/snp_enclaves/blobs/kernel_enabled_gpu/nvidia.ko
+COPY --from=apt_extract /blob /opt/fortanix/enclave-os/blobs
+COPY init /opt/fortanix/enclave-os/blobs/init
+COPY kernel_disabled_gpu/.config /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/bzImage.config
+COPY kernel_disabled_gpu/bzImage /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/bzImage
+COPY kernel_enabled_gpu/.config /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/bzImage.config
+COPY kernel_enabled_gpu/bzImage /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/bzImage
+COPY kernel_enabled_gpu/nvidia-uvm.ko /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/nvidia-uvm.ko
+COPY kernel_enabled_gpu/nvidia.ko /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/nvidia.ko
 
 ENTRYPOINT [ "./server" ]
--- a/docker/snp/parent-base/Dockerfile
+++ b/docker/snp/parent-base/Dockerfile
@@ -124,7 +124,7 @@ RUN apt-get update \
 COPY --from=qemu-builder /opt/qemu /opt/qemu
 ENV PATH="/opt/qemu/bin:${PATH}"
 
-COPY --from=parent-base /etc/dnsmasq.d /etc/dnsmasq.d
+COPY --from=parent-base dnsmasq.d /etc/dnsmasq.d
 
 # Fail-fast sanity check.
 RUN qemu-system-x86_64 --version
--- a/docker/snp/parent-base/Dockerfile
+++ b/docker/snp/parent-base/Dockerfile
@@ -124,7 +124,7 @@ RUN apt-get update \
 COPY --from=qemu-builder /opt/qemu /opt/qemu
 ENV PATH="/opt/qemu/bin:${PATH}"
 
-COPY --from=parent-base dnsmasq.d /etc/dnsmasq.d
+COPY --from=parent-base /etc/dnsmasq.d /etc/dnsmasq.d
 
 # Fail-fast sanity check.
 RUN qemu-system-x86_64 --version
--- a/tools/container-converter/Cargo.lock
+++ b/tools/container-converter/Cargo.lock
@@ -145,6 +145,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
+name = "base64"
+version = "0.21.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
+
+[[package]]
 name = "bindgen"
 version = "0.59.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -333,6 +339,7 @@ dependencies = [
  "docker-image-reference",
  "env_logger",
  "exitcode",
+ "fortanix-vme-initramfs",
  "futures",
  "http",
  "lazy_static",
@@ -349,6 +356,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "toml",
+ "ureq",
  "url",
 ]
 
@@ -357,6 +365,12 @@ name = "core-foundation-sys"
 version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5827cebf4670468b8772dd191856768aedcb1b0278a04f989f7766351917b9dc"
+
+[[package]]
+name = "cpio"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "27e77cfc4543efb4837662cb7cd53464ae66f0fd5c708d71e0f338b1c11d62d3"
 
 [[package]]
 name = "crc32fast"
@@ -405,6 +419,17 @@ name = "cxxbridge-macro"
 version = "1.0.68"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34fa7e395dc1c001083c7eed28c8f0f0b5a225610f3b6284675f444af6fab86b"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
+name = "derivative"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fcc3dd5e9e9c0b295d6e1e4d811fb6f157d5ffd784b8d202fc62eac8035a770b"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -537,6 +562,18 @@ checksum = "5fc25a87fa4fd2094bffb06925852034d90a17f0d1e05197d4956d3555752191"
 dependencies = [
  "matches",
  "percent-encoding",
+]
+
+[[package]]
+name = "fortanix-vme-initramfs"
+version = "0.1.0"
+source = "git+https://github.com/fortanix/rust-sgx.git#3fe3bf30f896e855302be25bf5af913a3b13b701"
+dependencies = [
+ "cpio",
+ "derivative",
+ "flate2",
+ "normalize-path",
+ "thiserror",
 ]
 
 [[package]]
@@ -1016,6 +1053,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "normalize-path"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f5438dd2b2ff4c6df6e1ce22d825ed2fa93ee2922235cc45186991717f0a892d"
+
+[[package]]
 name = "num-integer"
 version = "0.1.45"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1311,6 +1354,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "ring"
+version = "0.16.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3053cf52e236a3ed746dfc745aa9cacf1b791d846bdaf412f60a8d7d6e17c8fc"
+dependencies = [
+ "cc",
+ "libc",
+ "once_cell",
+ "spin",
+ "untrusted",
+ "web-sys",
+ "winapi",
+]
+
+[[package]]
 name = "rustc-hash"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1321,6 +1379,28 @@ name = "rustc-serialize"
 version = "0.3.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fe834bc780604f4674073badbad26d7219cadfb4a2275802db12cbae17498401"
+
+[[package]]
+name = "rustls"
+version = "0.21.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cd8d6c9f025a446bc4d18ad9632e69aec8f287aa84499ee335599fabd20c3fd8"
+dependencies = [
+ "log",
+ "ring",
+ "rustls-webpki",
+ "sct",
+]
+
+[[package]]
+name = "rustls-webpki"
+version = "0.101.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3c7d5dece342910d9ba34d259310cae3e0154b873b35408b787b59bce53d34fe"
+dependencies = [
+ "ring",
+ "untrusted",
+]
 
 [[package]]
 name = "ryu"
@@ -1346,6 +1426,16 @@ name = "scratch"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9c8132065adcfd6e02db789d9285a0deb2f3fcb04002865ab67d5fb103533898"
+
+[[package]]
+name = "sct"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d53dcdb7c9f8158937a7981b48accfd39a43af418591a5d008c7b22b5e1b7ca4"
+dependencies = [
+ "ring",
+ "untrusted",
+]
 
 [[package]]
 name = "serde"
@@ -1395,7 +1485,7 @@ name = "shiplift"
 version = "0.7.0"
 source = "git+https://github.com/fortanix/shiplift.git#07ee2d8d19bb8435d6c4fb38c0b5a2a905f23486"
 dependencies = [
- "base64",
+ "base64 0.13.0",
  "byteorder",
  "bytes 1.1.0",
  "chrono",
@@ -1474,6 +1564,12 @@ dependencies = [
  "libc",
  "winapi",
 ]
+
+[[package]]
+name = "spin"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6e63cff320ae2c57904679ba7cb63280a3dc4613885beafb148ee7bf9aa9042d"
 
 [[package]]
 name = "static_assertions"
@@ -1775,6 +1871,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "957e51f3646910546462e67d5f7599b9e4fb8acdd304b087a6494730f9eebf04"
 
 [[package]]
+name = "untrusted"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a156c684c91ea7d62626509bce3cb4e1d9ed5c4d978f7b4352658f96a4c26b4a"
+
+[[package]]
+name = "ureq"
+version = "2.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f8cdd25c339e200129fe4de81451814e5228c9b771d57378817d6117cc2b3f97"
+dependencies = [
+ "base64 0.21.7",
+ "flate2",
+ "log",
+ "once_cell",
+ "rustls",
+ "rustls-webpki",
+ "url",
+ "webpki-roots",
+]
+
+[[package]]
 name = "url"
 version = "2.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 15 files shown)


docker/snp/parent-base/Dockerfile
```diff
@@ -124,7 +124,7 @@ RUN apt-get update \
 COPY --from=qemu-builder /opt/qemu /opt/qemu
 ENV PATH="/opt/qemu/bin:${PATH}"
 
-COPY --from=parent-base dnsmasq.d /etc/dnsmasq.d
+COPY --from=parent-base /etc/dnsmasq.d /etc/dnsmasq.d
```
anubhavg64: CI is failing at `docker/snp/parent-base/Dockerfile:127` with `"/etc/dnsmasq.d": not found.` 
I'm trying to understand why we changed this.

sardok: Because the way it is, was not working for me. Letme revert it if its breaking CI scripts.

docker/snp/Dockerfile
```diff
@@ -15,18 +15,24 @@ cp ovmf_dir/usr/share/ovmf/OVMF.amdsev.fd /blob
...
 
-COPY --from=apt_extract /blob /usr/share/snp_enclaves/blobs
-COPY kernel_disabled_gpu/.config /usr/share/snp_enclaves/blobs/kernel_disabled_gpu/bzImage.config
-COPY kernel_disabled_gpu/bzImage /usr/share/snp_enclaves/blobs/kernel_disabled_gpu/bzImage
-COPY kernel_enabled_gpu/.config /usr/share/snp_enclaves/blobs/kernel_enabled_gpu/bzImage.config
-COPY kernel_enabled_gpu/bzImage /usr/share/snp_enclaves/blobs/kernel_enabled_gpu/bzImage
-COPY kernel_enabled_gpu/nvidia-uvm.ko /usr/share/snp_enclaves/blobs/kernel_enabled_gpu/nvidia-uvm.ko
-COPY kernel_enabled_gpu/nvidia.ko /usr/share/snp_enclaves/blobs/kernel_enabled_gpu/nvidia.ko
+COPY --from=apt_extract /blob /opt/fortanix/enclave-os/blobs
+COPY init /opt/fortanix/enclave-os/blobs/init
```
rui-alm: I'm currently aligning the two versions of init script, with and without GPU support, with how the kernel is being packaged: https://jenkins.fortanix.net/job/build-salmiac-kernel/52/

Does it make sense to you to copy the right version from the specific directories?

sardok: Pushed a fix here: https://github.com/fortanix/salmiac/pull/72/changes/392b35af3856236bc2aecd7521af2f2dd39cef08

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,245 @@
...
+    let run_cmd = GenericEnclaveImageBuilder::enclave_command_string(
+        enclave_settings,
+        &Path::new(INSTALLATION_DIR),
+        "enclave",
+    );
+
+    let mut blobs_dir = PathBuf::from(format!("{}/blobs", INSTALLATION_DIR));
+    let init = {
+        if enclave_settings.gpu_passthrough {
+            blobs_dir.push(BLOBS_SUBDIR_GPU_ENABLED);
```
tvsfx: use `Path::join` so we don't need to mutate `blobs_dir`

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,245 @@
...
+    let mut fs_tree = FsTree::new()
+        .add_file("env", Cursor::new(enclave_settings.env_vars.join("\n")))
+        .add_file("cmd", Cursor::new(run_cmd))
+        .add_executable("init", Cursor::new(init));
+
+    // Add enclave-base filesystem to rootfs
+    info!("Adding enclave-base filesystem to initramfs rootfs...");
+    fs_tree = add_enclave_base_to_initramfs(fs_tree, &mut enclave_base_archive)?;
+
+    // Add basic rootfs structure that `init` expects regardless of enclave-base has it or not
```
tvsfx: nit: language
> regardless of enclave-base has it or not

tools/container-converter/src/image_builder/parent/snp.rs
```diff
@@ -4,21 +4,274 @@
...
+            cmd: None,
+            entrypoint: Some(vec![
+                run_parent_cmd,
+                "--platform".to_string(),
+                "snp".to_string(),
+            ]),
+        }
+    }
+
+    fn append_start_enclave_command(
```
tvsfx: Can we move these duplicated functions to `parent/mod.rs`?

tools/container-converter/src/resources/parent/start-parent.sh
```diff
@@ -1,17 +1,58 @@
...
+
+if [ "$PLATFORM" == "nitro" ]; then
+    # Setup nitro environment
+    source /etc/profile.d/nitro-cli-env.sh
+
+    # Check if nitro-cli is properly installed
+    nitro-cli --version
+elif [ "$PLATFORM" == "snp" ]; then
+    # Check if qemu is properly installed
+    qemu-system-x86_64 --version
```
tvsfx, line 54: this won't do much without `set -e`

tools/container-converter/src/lib.rs
```diff
@@ -170,7 +177,24 @@ async fn run0(
 
     info!("Building enclave image!");
     let image_result = {
-        let enclave_base_image_str = env::var("ENCLAVE_IMAGE").unwrap_or(ENCLAVE_IMAGE.to_string());
+        let enclave_base_image_str = env::var("ENCLAVE_IMAGE").unwrap_or_else(|_| {
+            #[cfg(platform = "nitro")]
```
tvsfx: wouldn't it be easier to have an `enclave_image_env_var` function that is platform specific in `container-converter/{snp,nitro}.rs`

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,245 @@
...
+                ));
+            }
+        };
+        mode |= file_type.bits() as u32;
+
+        let entry_type = header.entry_type();
+        fs_tree = if entry_type.is_dir() {
+            fs_tree.add_directory_with_permissions(&target_path, mode)
+        } else if entry_type.is_symlink() || entry_type.is_hard_link() {
+            // TODO: For now we're treating a hardlink as an softlink.
```
tvsfx, line 249: I'm a bit scared of hard-to-foresee breakage this might cause when applied to an entire os userspace. Could we at least do a test-run to see what files are affected?

Discussed with @sardok offline, and we have the impression that `cpio-rs` is ill-suited to supporting hardlinks, because [it overwrites](https://docs.rs/cpio/0.4.1/src/cpio/lib.rs.html#31) ino's at CPIO write time. Alternatively, we could try to use the `cpio` cli tool.

tvsfx: https://github.com/jcreekmore/cpio-rs/issues/23

tools/container-converter/src/image_builder/enclave/snp.rs
sardok: Fixed: https://github.com/fortanix/salmiac/pull/72/changes/20c1df03fd942ae9675a3b51a873ecb8151bab19

tools/container-converter/src/image_builder/parent/snp.rs
sardok: Fixed https://github.com/fortanix/salmiac/pull/72/changes/20c1df03fd942ae9675a3b51a873ecb8151bab19

tools/container-converter/src/resources/parent/start-parent.sh
sardok: Fixed: https://github.com/fortanix/salmiac/pull/72/changes/20c1df03fd942ae9675a3b51a873ecb8151bab19

tools/container-converter/src/lib.rs
sardok: I added them to `PlatformEnclaveImageBuilder` function: https://github.com/fortanix/salmiac/pull/72/changes/d68293cf31cc124734c0da4dc958c6c245fc20ad. Let'me know how it looks now. I think, we should move this base image getting functionality from `lib.rs` to something like `PlatformEnclaveImageBuilder::new`. IMO, `lib.rs` is setting up things more than it is supposed to.

tools/container-converter/src/image_builder/enclave/snp.rs
sardok: Ok, i've added a warning when hardlink is observed. So we can see them during the conversion. See: https://github.com/fortanix/salmiac/pull/72/changes/a5293aa828162eb8c10c3493489fe365f4083aae

Here are the hardlinks we have in enclave base images;

```
enclave-base:
usr/bin/perl5.38.2
usr/bin/uncompress
usr/lib/klibc/bin/gzip
usr/lib/klibc/bin/poweroff
usr/lib/klibc/bin/reboot
usr/lib/klibc/bin/zcat
```

```
enclave-base-snp-gpu:
usr/bin/perl5.38.2
usr/bin/uncompress
```
Checked `usr/bin/uncompress` it is part of `gzip` and shares the same inode with `/usr/bin/gunzip`.

The hardlink issue we're having is annoying but i can't think about a scenario where symlink does not do the job except removal of hardlink/target file (which i think very unlikely). I am testing the snp image invocation now. If hardlink turns out to be necessary, i can switch back to cpio tools also it requires some effort as i briefly mentioned that it was not straightforward to use.

tools/container-converter/src/image_builder/enclave/snp.rs
sardok: I use many `push/pop` calls of `PathBuf` in this function. There are bunch of subdirs/files, needed to be joined. So `push/pop` are more convenient to use for me, also probably `push/pop` causes less allocation then `.join` (maybe none)..

tvsfx:
```rust
        blobs_dir.pop(); // gpu enabled/disabled subdir
        blobs_dir.pop(); // blobs dir
```
The fact that you (feel you) need the comments here says enough about how readable this code is, I think. I wouldnt worry about the single allocation for a one-off function call; it's more idiomatic (less bug-prone) to keep `blobs_dir` immutable, or even a constant.
(The same argument might be true for some other uses of `pop` in this PR)

Also, just to clarify: Im not against `.push`, its mostly the `.pop` call that makes things less readable if we just want to reuse the original, unaltered path elsewhere.

docker/snp/Dockerfile
```diff
@@ -15,18 +15,29 @@ cp ovmf_dir/usr/share/ovmf/OVMF.amdsev.fd /blob
 
 FROM ubuntu:24.04
 
-RUN mkdir -p /var/log/snp_enclaves
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && \
+    apt-get install -y cryptsetup-bin
+
+RUN mkdir -p /var/log/enclave-os
```
tvsfx, line 23: What do we use this log folder for? I can't immediately find it

docker/snp/Dockerfile
```diff
@@ -15,18 +15,29 @@ cp ovmf_dir/usr/share/ovmf/OVMF.amdsev.fd /blob
...
-COPY kernel_enabled_gpu/nvidia.ko /usr/share/snp_enclaves/blobs/kernel_enabled_gpu/nvidia.ko
+COPY --from=apt_extract /blob /opt/fortanix/enclave-os/blobs
+
+# Copy gpu disabled artifacts
+COPY kernel_disabled_gpu/init /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/init
+COPY kernel_disabled_gpu/.config /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/bzImage.config
+COPY kernel_disabled_gpu/bzImage /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/bzImage
+
+# Copy gpu enabled artifacts
+COPY kernel_enabled_gpu/init /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/init
```
tvsfx, line 37: TODO (before merge): add missing drivers

I don't think we need to hardcode the list of drivers in the code anymore, given that we choose here already what drivers we will COPY in to the dockerfile?

tools/container-converter/src/image_builder/enclave/mod.rs
```diff
@@ -24,6 +24,10 @@ use std::path::Path;
 
 use api_model::converter::{CertificateConfig, ConverterOptions, DsmConfiguration};
 use api_model::enclave::{CcmBackendUrl, EnclaveManifest, FileSystemConfig, UserConfig};
+#[cfg(platform = "nitro")]
```
tvsfx, line 27: I think it makes more sense to put this cfg block in `api_model/lib.rs`, where the types are defined.

tools/container-converter/src/image_builder/enclave/mod.rs
```diff
@@ -100,6 +108,8 @@ impl EnclaveSettings {
                 .dsm_configuration
                 .clone()
                 .unwrap_or_default(),
+            #[cfg(platform = "snp")]
+            gpu_passthrough: _enclaves_options.enable_gpu_passthrough.unwrap_or(false),
```
tvsfx, line 112: Nit: this is technically not idiomatic rust and will trip up this lint: https://rust-lang.github.io/rust-clippy/master/index.html#used_underscore_binding. You could #[allow(clippy::used_underscore_binding)] on this function. Or (Imo better) make the arg platform specific until we need it elsewhere. Same is true for `get_enclave_base_image_name`.

tools/container-converter/src/image_builder/enclave/mod.rs
```diff
@@ -141,6 +151,25 @@ impl<'a> EnclaveImageBuilder<'a> {
         &self,
         docker_util: &dyn DockerUtil,
         archive_path: &Path,
+    ) -> Result<Archive<File>> {
+        self.export_docker_image_file_system(docker_util, archive_path, self.client_image_reference)
+            .await
+    }
+
+    pub(crate) async fn export_enclave_base_file_system(
```
tvsfx, line 159: function unused in `nitro` mode.

tools/container-converter/build.rs
```diff
@@ -23,6 +23,11 @@ const RESOURCES_PARENT_DIR: &str = "src/resources/parent";
 const RESOURCES_ENCLAVE_DIR: &str = "src/resources/enclave";
 
 fn main() -> Result<(), Box<dyn Error>> {
+    println!("cargo:rerun-if-changed=../../vsock-proxy/src");
```
tvsfx, line 26: `vsock-proxy/src` does not exist; this would cause `build.rs` to run unconditionally

tools/container-converter/build.rs
```diff
@@ -53,11 +58,13 @@ fn main() -> Result<(), Box<dyn Error>> {
     let current_dir = env::current_dir().expect("Failed retrieving current directory");
 
     let status = {
+        let platform =
+            env::var(SALMIAC_PLATFORM_ENV).unwrap_or_else(|_| SALMIAC_PLATFORM.to_string());
```
tvsfx, line 62: [Not your code, but since this got changed Ill say it anyway] This fallback is odd to me; if `SALMIAC_PLATFORM_ENV` is not set, then `emit_cfg_from_env` above [will produce an error anyway](https://github.com/fortanix/salmiac/blob/f02d858116b86e6bd0dffa9263283d485a03049c/build-support/src/lib.rs?plain=1#L19). If we want a fallback we should imo implement it centrally in `build-support`.

tools/container-converter/Cargo.lock
```diff
@@ -37,9 +37,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.57"
+version = "1.0.102"
```
tvsfx, line 40: did you run `cargo update` unconditionally? I can't seem to figure out why `anyhow` got updated, and if I do 

```sh
git restore --worktree --source 61d19c9e41f9bac271eca8ba79818753bc5a459e -- tools/container-converter/Cargo.lock
cd tools/container-converter/
SALMIAC_PLATFORM=snp cargo build
git diff HEAD -- :/tools/container-converter/Cargo.lock
```

then my build succeeds with older versions of some of these packages

tools/container-converter/src/image_builder/enclave/mod.rs
tvsfx: Getting the following warnings in `snp` mode :

```sh
warning: method `create_resource` is never used
  --> src/file.rs:39:19
   |
22 | impl BuildContext {
   | ----------------- method in this implementation
...
39 |     pub(crate) fn create_resource(&self, resource: Resource) -> Result<(), String> {
   |                   ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default

warning: associated items `IMAGE_COPY_DEPENDENCIES`, `create_manifest_file`, `enclave_image`, `retag_client_image`, `create_requisites`, and `docker_file_contents` are never used
   --> src/image_builder/enclave/mod.rs:147:22
    |
125 | impl<'a> EnclaveImageBuilder<'a> {
    | -------------------------------- associated items in this implementation
...
147 |     pub(crate) const IMAGE_COPY_DEPENDENCIES: &'static [&'static str] =
    |                      ^^^^^^^^^^^^^^^^^^^^^^^
...
209 |     pub(crate) fn create_manifest_file(
    |                   ^^^^^^^^^^^^^^^^^^^^
...
261 |     pub(crate) fn enclave_image(&self) -> String {
    |                   ^^^^^^^^^^^^^
...
265 |     fn retag_client_image(&self, tag: &str) -> String {
    |        ^^^^^^^^^^^^^^^^^^
...
275 |     pub(crate) fn create_requisites(
    |                   ^^^^^^^^^^^^^^^^^
...
323 |     fn docker_file_contents(&self, mut enclave_settings: EnclaveSettings) -> DockerFile {
    |        ^^^^^^^^^^^^^^^^^^^^

```

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+
+use crate::image_builder::enclave::EnclaveImageBuilder as GenericEnclaveImageBuilder;
+use crate::image_builder::enclave::EnclaveSettings;
+use crate::image_builder::parent::snp::{BLOBS_SUBDIR_GPU_DISABLED, BLOBS_SUBDIR_GPU_ENABLED};
+use crate::image_builder::INSTALLATION_DIR;
+use crate::{ConverterError, ConverterErrorKind, Result};
 
 pub(crate) struct EnclaveImageBuilder<'a> {
-    pub(crate) enclave_image_builder: crate::image_builder::enclave::EnclaveImageBuilder<'a>,
+    pub(crate) enclave_image_builder: GenericEnclaveImageBuilder<'a>,
```
tvsfx, line 26: perhaps we should've just renamed the struct

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+
 impl<'a> EnclaveImageBuilder<'a> {
-    // TODO: RTE-941
+    const INIT_BIN: &'static str = "init";
+    const GPU_MODULES: &'static [&'static str] = &["nvidia.ko", "nvidia-uvm.ko"];
+    pub const INITRAMFS_FILENAME: &'static str = "initramfs.gz";
+
     pub(crate) async fn create_image(
         &self,
         input_repository: &crate::docker::DockerDaemon,
```
tvsfx, line 36: This should be `docker_util: &dyn DockerUtil,` for consistency and testing (dependency injection). We should probably make this generic in a separate PR.

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
-        enclave_settings: crate::image_builder::enclave::EnclaveSettings,
-        user_config: api_model::enclave::UserConfig,
-        image_env_vars: Vec<String>,
-        sender: std::sync::mpsc::Sender<crate::image::ImageToClean>,
+        enclave_settings: EnclaveSettings,
+        user_config: UserConfig,
+        env_vars: Vec<String>,
+        _sender: std::sync::mpsc::Sender<crate::image::ImageToClean>,
```
tvsfx, line 40: What is the `clean_docker_images` task for? It does not seem to be used for `snp`; should it be, or should we refactor the code?

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+            dsm_configuration,
+        };
+
+        let enclave_manifest_data =
+            serde_json::to_vec(&enclave_manifest).map_err(|err| ConverterError {
+                message: format!("Failed serializing enclave settings file. {:?}", err),
+                kind: ConverterErrorKind::RequisitesCreation,
+            })?;
+
+        let enclave_base_tar_path = work_dir.join("enclave-base.tar");
```
tvsfx, line 72: nit: make this a constant for consistency?

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+            &enclave_settings,
+            &enclave_manifest_data,
+        )
+        .map_err(|e| ConverterError {
+            message: format!("Failed to create initramfs: {}", e),
+            kind: ConverterErrorKind::EnclaveImageCreation,
+        })?;
+
+        Ok(SNPEnclavesMeasurements {
+            launch_measurement: HexString::new(vec![0; 48]),
```
tvsfx, line 94: add TODO for measurement?

tools/container-converter/src/image_builder/enclave/snp.rs
tvsfx: Also, nice use of join 😉

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+            }
+        })
     }
 }
+
+fn create_initramfs(
+    output_path: &Path,
+    mut enclave_base_archive: Archive<fs::File>,
+    enclave_settings: &EnclaveSettings,
+    enclave_manifest_data: &[u8],
```
tvsfx, line 115: this can be `Vec<u8>` to save yourself a clone

or even just `&EnclaveManifest` and do the serialization internally

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+    let mut fs_tree = FsTree::new()
+        .add_file("env", Cursor::new(get_env_vars(&enclave_settings.env_vars)))
+        .add_file("cmd", Cursor::new(run_cmd))
+        .add_executable("init", Cursor::new(init));
+
+    // Add enclave-base filesystem to rootfs
+    info!("Adding enclave-base filesystem to initramfs rootfs...");
+    fs_tree = add_enclave_base_to_initramfs(fs_tree, &mut enclave_base_archive)?;
+
+    // Set up minimal runtime directories. Do not create directories that exist
```
tvsfx, line 143: I was a bit confused by what you wanted to say with this comment; if the original enclave_base image already contained `rootfs/dev` then adding it below would be a noop, right? Also, unless Im misunderstanding, I dont think pre-existing dirs can interfere with visibility of the rootfs, since the moved mount will shadow the old dirs?

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+    info!("Building initramfs, it may take a while depending on size of it...");
+    let output_file = fs::File::create(output_path)?;
+    Initramfs::from_fs_tree(fs_tree, output_file).map_err(|e| {
+        IoError::new(
+            IoErrorKind::Other,
+            format!("Failed to build initramfs: {:?}", e),
+        )
+    })?;
+
+    info!("Built initramfs.gz at {}", output_path.display());
```
tvsfx, line 198: use `INITRAMFS_FILENAME`

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+                    IoErrorKind::Other,
+                    format!("Unsupported tar entry: {}", path_str),
+                ));
+            }
+        };
+        mode |= file_type.bits() as u32;
+
+        let entry_type = header.entry_type();
+        fs_tree = if entry_type.is_dir() {
+            fs_tree.add_directory_with_permissions(&target_path, mode)
```
tvsfx, line 247: why do these `fs_tree` methods take a `&str` again?  wouldnt `&Path` or `AsRef<Path>` have made our lives easier? Then we could've just called `join/push` once

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+}
+
+fn add_enclave_base_to_initramfs(
+    mut fs_tree: FsTree,
+    enclave_base_archive: &mut tar::Archive<fs::File>,
+) -> std::result::Result<FsTree, IoError> {
+    let entries = enclave_base_archive.entries()?;
+    for entry in entries {
+        let mut entry = entry?;
+        let path = entry.path()?.to_path_buf();
```
tvsfx, line 209:
```diff
-        let path = entry.path()?.to_path_buf();
+        let path = entry.path()?;
```

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+        let path_str = path.to_str().ok_or(IoError::new(
+            IoErrorKind::Other,
+            format!("Invalid path in archive: {:?}", path),
```
tvsfx, line 212: style: start errors in small letters.

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+        ))?;
+
+        let target_path = format!("rootfs/{}", path_str);
+
+        let header = entry.header();
+        // Some tar archives doesn't preserve modes.
```
tvsfx, line 218: was this a problem for you when implementing? If not we might just want to bail out, since the docs say the `mode` is only unset if it's corrupted?

tools/container-converter/src/image_builder/enclave/snp.rs
```diff
@@ -4,23 +4,283 @@
...
+                return Err(IoError::new(
+                    IoErrorKind::Other,
+                    format!("Unsupported tar entry: {}", path_str),
+                ));
+            }
+        };
+        mode |= file_type.bits() as u32;
+
+        let entry_type = header.entry_type();
+        fs_tree = if entry_type.is_dir() {
```
tvsfx, line 246: I think it would read more cleanly if you merged this switch on `entry_type` with the `match` above; you can use `let get_mode = |file_type| mode |= file_type.bits() as u32` to get the mode in each branch.

tools/container-converter/src/resources/parent/start-parent.sh
```diff
@@ -1,17 +1,58 @@
...
+            ;;
+        *)
+            break
+            ;;
+    esac
+done
+
+if [ "$PLATFORM" == "nitro" ]; then
+    # Setup nitro environment
+    source /etc/profile.d/nitro-cli-env.sh
```
pavankalyan0424, line 48: We can remove this line. As part of this commit - https://github.com/fortanix/salmiac/commit/2cc6df4b142ec06e3a0e6815d205f5a9a3063fcf#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddcR9-L18, We removed copying of profile.d dir. 
As we added -e to the script, and since the nitro-cli-env.sh is not getting copied, We are facing issue while running CI.

tools/container-converter/build.rs
sardok: This part has been updated.

docker/snp/Dockerfile
sardok: Ok, i guess it stems from Nitro part where `/var/log/nitro_enclaves` is used.

docker/snp/Dockerfile
sardok: Added in beta branch.

tools/container-converter/src/image_builder/enclave/snp.rs
sardok: I think the problem is here: https://github.com/aws/aws-nitro-enclaves-sdk-bootstrap/blob/f718dea60a9d9bb8b8682fd852ad793912f3c5db/init/init.c#L434.
Consider this initramfs structure:
```
/bin
/tmp
/dev
/rootfs/bin
/rootfs/dev
```
When `mount /rootfs / --move` is executed. The rootfs' `/bin` & `/dev` will be shadowed by the -before mount- `/bin` & `/dev`. For `rootfs/dev` its ok, it empty but `rootfs/bin` contains some symlinks to `rootfs/sbin`.

tools/container-converter/src/image_builder/enclave/snp.rs
sardok: Implemented in v0.2

tools/container-converter/src/image_builder/enclave/snp.rs
sardok: You're right. I was confused here. There is no need for those directories either. Its probably best to not make any assumptions about enclave-base fs structure. I remove them in the follow-up PR.

sardok:
The comments left to this PR are addressed to https://github.com/fortanix/salmiac/pull/82. Closing as development will continue on `beta` branch.

PR #73, opened 2026-05-15
[RTE-919] Add serde rename to match with api-model

pavankalyan0424:
Added serde field rename for backward compatibility and match with api-model


tvsfx:
~~Can you squash these commits please?~~
I guess it's fine since we squash merge by default in Salmiac, iirc. (the squash commit message still gets polluted, but that's less annoying)

tvsfx (approved):

PR #76, opened 2026-05-27
[RTE-964] Use tar files as fallbacks for base images

pavankalyan0424:
This PR implements a fallback mechanism to load the image from tar file when it is not available in the local Docker repository.


rui-alm (changes requested):
It is not aligned with the split done in https://github.com/fortanix/salmiac/pull/65.
For example, the parent images are now tagged parent-image-{platform}, which doesn't work with what the code currently expects, which is just `parent-image`.

PR #79, opened 2026-06-03
RTE-987: ENCLAVEOS_DISABLE_DEFAULT_CERTIFICATE should accept bool values properly

dlegalov-fortanix:
The `ENCLAVEOS_DISABLE_DEFAULT_CERTIFICATE` environmental variable would disable the default cert generation if set to
```bash
ENCLAVEOS_DISABLE_DEFAULT_CERTIFICATE=false
```
due to _any_ value being interpreted to disable the default cert.
This change makes the variable behave in a more logical manner.

Ported to `beta` in this PR: https://github.com/fortanix/salmiac/pull/80

### DIFF
--- a/vsock-proxy/enclave/src/enclave.rs
+++ b/vsock-proxy/enclave/src/enclave.rs
@@ -189,7 +189,8 @@ pub(crate) async fn run(
     {
         // Check that the value is true or zero
         let val = val.trim(); // Remove any whitespace
-        if let Ok(_i @ 0) = val.parse::<i64>() {
+        if let Ok(_i @ 1..) = val.parse::<i64>() {
+            // Values of 1 or greater are considered logically true
             true
         } else if val.to_lowercase() == "true" {
             true
--- a/vsock-proxy/enclave/src/enclave.rs
+++ b/vsock-proxy/enclave/src/enclave.rs
@@ -182,11 +182,23 @@ pub(crate) async fn run(
     // execution and it is considered an error if they do.
     let mut background_tasks = start_background_tasks(networking_setup_result.tap_devices);
 
-    let skip_def_cert_req = setup_result
+    let skip_def_cert_req = if let Some((_, val)) = setup_result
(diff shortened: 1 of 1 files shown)


vsock-proxy/enclave/src/enclave.rs
```diff
@@ -182,11 +182,23 @@ pub(crate) async fn run(
...
+    let skip_def_cert_req = if let Some((_, val)) = setup_result
         .env_vars
         .iter()
-        .find(|(k, v)| k == "ENCLAVEOS_DISABLE_DEFAULT_CERTIFICATE" && !v.trim().is_empty())
-        .is_some();
+        .find(|(k, _)| k == "ENCLAVEOS_DISABLE_DEFAULT_CERTIFICATE")
+    {
+        // Check that the value is true or zero
+        let val = val.trim(); // Remove any whitespace
+        if let Ok(_i @ 0) = val.parse::<i64>() {
```
sardok: I think `true` & `1` (or anything non-zero) should have the same effect. Here it seems to me `ENCLAVEOS_DISABLE_DEFAULT_CERTIFICATE=0` & `ENCLAVEOS_DISABLE_DEFAULT_CERTIFICATE=true` is setting to `skip_def_cert_req`. Probably this line should be `if let Ok(_i @ 1..) = val.parse<i64>()`

dlegalov-fortanix: You're right @sardok , I will adjust the logic, I just checked the way Bash propagates env vars; I was mixing up return codes with the traditional meaning of booleans.
I will set 1  or greater to mean true, 0 or less to mean false.

dlegalov-fortanix: Resolved with commit https://github.com/fortanix/salmiac/pull/80/changes/817830b670428382196b3f2dcb8f8f2ef5dbba44

sardok (approved):

PR #81, opened 2026-06-03
[RTE-964] Use tar files as fallbacks for base images

pavankalyan0424:
This PR implements a fallback mechanism to load the image from tar file when it is not available in the local Docker repository or not set.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 /build/
 /build-support/target
 /artifacts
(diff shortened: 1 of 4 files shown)


sardok:
The base PR, won't be merged into master. Instead we'll keep development on `beta`. So i think this PR should re-target to beta.

rui-alm (approved):

PR #88, opened 2026-06-11
[RTE-1050] - Do not rely on docker cache to build converter image

rui-alm:


docker/snp/build-conv-container.sh
```diff
@@ -18,6 +18,5 @@ docker save -o ./staging/parent-base.tar parent-base-snp
 cp -r ../../tools/sev-snp-measure ./staging/sev-snp-measure
 
 # Build the converter
-docker build -t snp-converter .
-#docker build --no-cache -t snp-converter .
+docker build --no-cache -t snp-converter .
```
aditijannu, line 21: Do you think we should do this for all docker images built under `docker` - includes the base images and converters.

rui-alm: Yes, we should. That would guarantee that the latest packages at the time of build are the ones installed and that the right local files are the ones copied into the image.

pavankalyan0424 (approved):
LGTM.

sardok (approved):

PR #89, opened 2026-06-11
[RTE-947] Modify internal create_block_file function to use bytes instead of megabytes

alex-fortanix:


alex-fortanix:
I don't know the nuances of Salmiac well yet so I'm not sure how important, if at all, data packing is on the block files. Currently in the PR I am rounding up to the next megabyte for the block file size, which is the behavior that currently exists here. Will going down to byte or kilobyte precision cause any issues?

tools/container-converter/src/image_builder/enclave/mod.rs
```diff
@@ -442,6 +442,7 @@ impl<'a> GenericEnclaveImageBuilder<'a> {
         async fn get_available_disc_space(block_file_dir: &Path) -> Result<u64> {
             statfs(block_file_dir)
                 .map(|e| e.block_size() as u64 * e.blocks_available())
+                .map(|size_mb| size_mb * MEGA_BYTE)
```
rui-alm: Is this function now reporting 2 orders of magnitude more space than it should? The block size is in bytes and we're multiplying by 1MB.

rui-alm (changes requested):

tools/container-converter/src/image_builder/enclave/mod.rs
alex-fortanix: Doesn't it give us block size in megabytes? In the current code it is comparing this output to the image size in megabytes

alex-fortanix: @rui-alm nope you were right, the comparison in the current code is incorrect. It is comparing the size of the image in megabytes to the available disk space in bytes so it will probably always evaluate to true.

Removed the multiplier.

rui-alm (approved):

PR #93, opened 2026-06-15
[RTE-1002] Add vsock proxy tdx support.

sardok:
This PR introduces Tdx platform & adds vsock-proxy implementation of this new platform.  Some refactoring is made to maximize code re-use between snp, tdx and simulator platforms in vsock-proxy.

### DIFF
--- a/ENV_VARS.md
+++ b/ENV_VARS.md
@@ -55,6 +55,10 @@ registered as a debug build in a test-only deployment account.
   passed to the nitro-cli run command.
 
 ##### AMD SEV-SNP enclaves settings variables
+- CPU_COUNT - Override the `-smp` param passed while running the enclave i.e.
+  passed to the qemu run command.
+- MEM_SIZE - Override the `-m` param & memory backend object passed while running
+  the enclave i.e. passed to the qemu run command.
 - SNP_GPU_BDF - For use with EnclaveOS containers with GPU passthrough.
   Consists of the BDF (Bus/Device/Function) of the GPU to be used.
 - SNP_CPU - QEMU CPU type, defaults to `EPYC-v4,-tsa-sq-no,-tsa-l1-no,family=0,model=0,stepping=0`.
--- a/vsock-proxy/parent/src/platform/tdx.rs
+++ b/vsock-proxy/parent/src/platform/tdx.rs
@@ -22,10 +22,6 @@ impl TdxPlatform {
     }
 
     fn tdx_guest() -> String {
-        // At the moment the host does not run quote-generation service.
-        // With quote-generation service enablement, we can either connect to it via vsock or socket file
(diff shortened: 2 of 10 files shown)


vsock-proxy/parent/src/platform/tdx.rs
```diff
@@ -0,0 +1,109 @@
...
+
+struct TdxPlatform;
+
+impl TdxPlatform {
+    fn memory_backend() -> String {
+        format!("memory-backend-ram,id={MEMORY_BACKEND_ID},size={MEMORY_SIZE}")
+    }
+
+    fn tdx_guest() -> String {
+        // At the moment the host does not run quote-generation service.
```
gilanghamidy: We are not going to use the DCAP TDX quote generation as we use our own TDX attestation scheme. So this part/comment can be ignored.

vsock-proxy/parent/src/platform/tdx.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashSet;
+
+    use super::*;
+
+    #[test]
+    fn test_build_qemu_snp_args() {
```
gilanghamidy: Name should be `tdx` instead of `snp`

ENV_VARS.md
```diff
@@ -55,6 +55,10 @@ registered as a debug build in a test-only deployment account.
   passed to the nitro-cli run command.
 
 ##### AMD SEV-SNP enclaves settings variables
+- CPU_COUNT - Override the `-smp` param passed while running the enclave i.e.
```
gilanghamidy, line 58: Typo, should be: `-snp`

gilanghamidy (commented):
Mostly looks fine, just minor stuff from my side. I will rebase my WIP branch on top of this.

ENV_VARS.md
sardok: `-smp` is correct. It is the argument that qemu expects. I think this name is coming from linux smp.

vsock-proxy/parent/src/platform/tdx.rs
sardok: Out of curiosity, will our custom TDX attestation scheme require any enclave argument?

vsock-proxy/parent/src/platform/tdx.rs
sardok: Thanks. Fixed.

vsock-proxy/parent/src/platform/tdx.rs
gilanghamidy: It should not be as it behaves similarly to the SNP attestation. So the parameter should be the same as the SNP one.

ENV_VARS.md
gilanghamidy: Ah wait, I thought `-snp` here is our argument, but now I realize the `-smp` from QEMU means for number of processor. Sorry for the confusion lol

gilanghamidy (approved):

PR #95, opened 2026-06-16
RTE-1045: parent-base-image changes to support tdx conversion

n7377747:
Moved the parent-base Dockerfile to a common qemu folder
And removed dnsmasq.d as it can be passed to docker build using `--build-context`


docker/tdx/parent-base/Dockerfile
rui-alm, line 1: Unlike the QEMU for SNP, this QEMU version isn't being built from source. Can we reproduce this QEMU build? What guarantees do we have that the package doesn't get modified or upgraded on the kobuk repo?

sardok: This version is very low compared to what we're using for snp. I guess, the challenge is about keeping the qemu source/version on par with tdx-measure's Dockerfile. One thing we can do is, we add support for download qemu tarball directly to the tdx-measure's Dockerfile.

docker/tdx/parent-base/dnsmasq.d
```diff
@@ -0,0 +1,23 @@
+# If you want dnsmasq to listen for DHCP and DNS requests only on
```
aditijannu: Do we need a copy of this file? We can avoid duplicating this file and reuse the original file from nitro parent base directory by passing `--build-context` when building the image. Or we can also move this file to a shared directory.

aditijannu:
To test the tdx parent-base image build, include the build step in the github workflows here - https://github.com/fortanix/salmiac/blob/beta/.github/workflows/build-docker-images.yml#L27

build-snp-container-converter.sh
rui-alm: Apparently there are now scripts which build base images, so the issue from #88 applies here. Could you make `--no-cache` the default behavior, please?

rui-alm (changes requested):

build-snp-container-converter.sh
```diff
@@ -63,7 +63,7 @@ docker build ${build_flags} -t "enclave-base-snp-gpu" .
 popd >/dev/null || exit
 
 echo "----- Building parent-base-snp -----"
-pushd "${salmiac_dir}/docker/snp/parent-base" >/dev/null || exit
+pushd "${salmiac_dir}/docker/qemu/parent-base" >/dev/null || exit
 docker build ${build_flags} -t "parent-base" .
```
rui-alm: Are we missing the build context here, now that `dnsmasq.d` has been removed?

build-snp-container-converter.sh
rui-alm, line 3: Can you add `set -e`, please? Otherwise the script will continue despite errors.

docker/snp/build-conv-container.sh
```diff
@@ -11,12 +11,13 @@ mv ./staging/container-converter ./staging/server
 
 # Package the enclave and parent base images for the converter
 docker save -o ./staging/enclave-base.tar enclave-base
-docker save -o ./staging/enclave-base-gpu.tar enclave-base-snp-gpu
-docker save -o ./staging/parent-base.tar parent-base-snp
+docker save -o ./staging/enclave-base-gpu.tar enclave-base-gpu
+docker save -o ./staging/parent-base.tar parent-base-qemu
```
rui-alm: The `build-snp-container-converter.sh` script is tagging the parent base image for SNP as `parent-base-snp`.

For now I think it's fine to revert this change to `parent-base-snp`, so the SNP converter script still works, but we can revisit this for the TDX converter, which will likely share logic with the SNP scripts.

rui-alm (changes requested):

.github/workflows/build-docker-images.yml
```diff
@@ -27,10 +27,10 @@ jobs:
...
-          docker build docker/snp/enclave-base-gpu -t enclave-base-snp-gpu
+          docker build docker/enclave-base-gpu -t enclave-base-gpu
           docker build docker/nitro/parent-base -t parent-base-nitro
           docker build \
             --build-context parent-base=docker/nitro/parent-base \
-            -t parent-base-snp \
-            docker/snp/parent-base
-          docker rmi enclave-base enclave-base-snp-gpu parent-base-nitro parent-base-snp
+            -t parent-base-qemu \
+            docker/qemu/parent-base
```
rui-alm: Could you please also set `--no-cache` on all of these builds, so we ensure the GitHub workflows are building from scratch?

rui-alm (approved):

PR #96, opened 2026-06-16
[RTE-1051] Add check for SNP_GPU_BDF when GPU passthrough is enabled and fix enclave startup error handling

alex-fortanix:


alex-fortanix:
Logs on case of failure due to invalid/unavailable GPU - Program exits after the error log

```
alex.pitsikoulis@TRY-66495-gpu01:~$ docker run     --rm     --privileged     --name="alex-test"     --runtime="runc"     -e ENCLAVEOS_DISABLE_DEFAULT_CERTIFICATE=false     -e ENCLAVEOS_DEBUG=debug     -e RUST_LOG=debug     -e MEM_SIZE=4096M  -e SNP_GPU_BDF=0   hello-world-snp
QEMU emulator version 10.2.2
Copyright (c) 2003-2025 Fabrice Bellard and the QEMU Project developers
This is nbd-server version 3.26.1
[2026-06-24T14:33:17Z WARN  parent] RW_STORAGE_SIZE is not present. Setting read/write block file size to a default value of 268435456
[2026-06-24T14:33:17Z INFO  parent] enclave_extra_args is []
[2026-06-24T14:33:17Z INFO  parent::parent] CORRECT VERSION 4 hopefully final!
[2026-06-24T14:33:17Z INFO  parent::parent] Checking presence of overlayfs parent directory.
[2026-06-24T14:33:17Z INFO  parent::parent] Creating overlayfs directory where the rw encrypted blockfile would be created...
[2026-06-24T14:33:17Z INFO  parent::parent] Spawning enclave process.
[2026-06-24T14:33:17Z INFO  parent::parent] Awaiting confirmation from enclave.
[2026-06-24T14:33:17Z ERROR parent] Parent exits with failure: GPU PCI device not found at /sys/bus/pci/devices/0
```

alex-fortanix:
Output in case when GPU passthrough is enabled at conversion time but no GPU is specified
```
QEMU emulator version 10.2.2
Copyright (c) 2003-2025 Fabrice Bellard and the QEMU Project developers
This is nbd-server version 3.26.1
[2026-06-24T19:30:32Z WARN  parent] RW_STORAGE_SIZE is not present. Setting read/write block file size to a default value of 268435456
[2026-06-24T19:30:32Z INFO  parent] enclave_extra_args is []
[2026-06-24T19:30:32Z INFO  parent::parent] Checking presence of overlayfs parent directory.
[2026-06-24T19:30:32Z INFO  parent::parent] Creating overlayfs directory where the rw encrypted blockfile would be created...
[2026-06-24T19:30:32Z INFO  parent::parent] Spawning enclave process.
[2026-06-24T19:30:32Z INFO  parent::parent] Awaiting confirmation from enclave.
[2026-06-24T19:30:32Z ERROR parent] Parent exits with failure: GPU passthrough was enabled at conversion time but SNP_GPU_BDF env var is not set
```
Error is returned and program exits before qemu process is started

alex-fortanix:
Output in case of GPU resource busy:

```
[2026-06-29T22:10:28Z DEBUG shared] Running subprocess qemu-system-x86_64 ["-enable-kvm", "-m", "4096M", "-smp", "2", "-cpu", "EPYC-v4,-tsa-sq-no,-tsa-l1-no,family=0,model=0,stepping=0", "-nographic", "-monitor", "none", "-no-reboot", "-bios", "/opt/fortanix/enclave-os/OVMF.amdsev.fd", "-machine", "q35,confidential-guest-support=sev0,vmport=off,memory-backend=ram1", "-object", "memory-backend-memfd,id=ram1,size=4096M,share=true,prealloc=false", "-object", "sev-snp-guest,id=sev0,cbitpos=51,reduced-phys-bits=1,kernel-hashes=on,policy=0x20000", "-object", "iommufd,id=iommufd0", "-device", "pcie-root-port,id=pci.1,bus=pcie.0", "-device", "vfio-pci,host=0000:21:00.0,bus=pci.1,iommufd=iommufd0,romfile=", "-fw_cfg", "name=opt/ovmf/X-PciMmio64Mb,string=262144", "-device", "vhost-vsock-pci,guest-cid=3", "-kernel", "/opt/fortanix/enclave-os/bzImage", "-initrd", "/opt/fortanix/enclave-os/initramfs.gz", "-append", "console=ttyS0 rdinit=/init loglevel=7"].
qemu-system-x86_64: -device vfio-pci,host=0000:21:00.0,bus=pci.1,iommufd=iommufd0,romfile=: vfio 0000:21:00.0: error bind device fd=22 to iommufd=21: Device or resource busy
[2026-06-29T22:10:29Z ERROR parent] Parent exits with failure: Process exited with a negative return code. Output is: Output { status: ExitStatus(unix_wait_status(256)), stdout: "", stderr: "" }
```

Error is returned and program exits before qemu process is started

vsock-proxy/parent/src/platform/qemu.rs
```diff
@@ -55,6 +57,10 @@ pub(super) trait QemuPlatform {
         if let Some(gpu) = self.gpu() {
             require_vfio_device(&gpu)?;
             require_file("IOMMUFD device", constants::IOMMU_DEVICE_PATH)?;
+        } else if env::var(constants::ENABLE_GPU_PASSTHROUGH_ENV_VAR).is_ok_and(|v| v == "true") {
+            return Err(String::from(
+                "GPU passthrough was enabled at conversion time but SNP_GPU_BDF env var is not set",
```
rui-alm: For TDX this variable will be `TDX_GPU_BDF`, so we should get it from the `gpu()` function when logging.

rui-alm (changes requested):

vsock-proxy/parent/src/platform/qemu.rs
alex-fortanix: Added the env var name as a const on the trait, changed the trait default of `gpu()` to use that env var if specified and removed the individual impls, and changed the log to use this const

vsock-proxy/parent/src/parent.rs
```diff
@@ -100,10 +100,27 @@ pub(crate) async fn run(args: ParentConsoleArguments) -> Result<UserProgramExitS
...
+
+        join_res = enclave_process => {
+            match join_res {
+                Ok(Err(e)) => Err(e),
+                Err(e) => Err(format!("Enclave task panicked or was cancelled: {:?}", e)),
+                Ok(_) => Err(String::from("Enclave exited unexpectedly")),
+            }
+        }
+
+        _ = sleep(Duration::from_secs(60)) => Err(String::from("Parent timed out waiting for connection from enclave")),
```
rui-alm: @sardok I see that you're modifying this block of code on #104 as well, so I'm inviting you to have a look at these changes as well.

I'm guessing the current behavior is that the parent waits forever if the enclave never responds, so I think it makes sense to have the option to timeout.

However, I don't like unconfigurable timeouts. Even though 1 minute may be a reasonable waiting time, I'd prefer for it to be configurable, and maybe even disabling the timeout.

If we think the timeout is beyond the scope of this PR, which was mostly about GPU checks, we can create a new ticket to define the timeout behavior.

rui-alm (commented):
You may need to rebase and solve conflicts.

vsock-proxy/parent/src/parent.rs
sardok: I agree, even though, 1 min seems reasonable, if enclave can't make it in 1 min then whole converted image becomes unusable. There is a ticket about adding cli args to vsock-proxy to get rid of passing enclave options based on env variables, with that ticket, we can add timeout support as well. Until that, my vote would be removal of timeout.

tools/container-converter/src/image_builder/parent/snp.rs
```diff
@@ -137,8 +137,19 @@ impl<'a> ParentImageBuilder<'a> {
...
+                .unwrap_or_default(),
+        );
+
+        let env_vars = vec![
+            log_env,
+            cpu_count_env,
+            mem_size_env,
+            eos_debug_env,
+            gpu_passthrough_env,
+        ];
```
sardok: This file is refactored and latest version is available on `beta` now. This part seems to reqiure re-writing.

vsock-proxy/parent/src/parent.rs
alex-fortanix: Ok I have removed it for now

tools/container-converter/src/image_builder/parent/snp.rs
alex-fortanix: Rebased and thankfully this just resulted in me needing to remove a few lines. Tested and still working as expected

vsock-proxy/parent/src/parent.rs
```diff
@@ -100,10 +100,26 @@ pub(crate) async fn run(args: ParentConsoleArguments) -> Result<UserProgramExitS
     info!("Spawning enclave process.");
     let guest_launch_result = crate::platform::launch_guest();
     // todo: will be used in https://fortanix.atlassian.net/browse/SALM-300
-    let _enclave_process = guest_launch_result.enclave_process;
+    let enclave_process = guest_launch_result.enclave_process;
 
     info!("Awaiting confirmation from enclave.");
-    let mut enclave_port = create_vsock_stream(VSOCK_PARENT_PORT).await?;
+    let mut enclave_port = tokio::select! {
+        accept = create_vsock_stream(VSOCK_PARENT_PORT) => {
```
sardok: Not sure if you prefer explicit `match accept` but this can be single line, `accept = create_vsock_stream(VSOCK_PARENT_PORT) => accept,`.

vsock-proxy/parent/src/parent.rs
```diff
@@ -100,10 +100,26 @@ pub(crate) async fn run(args: ParentConsoleArguments) -> Result<UserProgramExitS
     info!("Spawning enclave process.");
     let guest_launch_result = crate::platform::launch_guest();
     // todo: will be used in https://fortanix.atlassian.net/browse/SALM-300
```
sardok: I think we can remove `// todo` now.

vsock-proxy/parent/src/parent.rs
alex-fortanix: Ah whoops good catch, this was left over from when I was trying to do something convoluted with the result. Simplified to what you mentioned

vsock-proxy/parent/src/parent.rs
alex-fortanix: removed

vsock-proxy/parent/src/platform/qemu.rs
```diff
@@ -55,6 +63,13 @@ pub(super) trait QemuPlatform {
         if let Some(gpu) = self.gpu() {
             require_vfio_device(&gpu)?;
             require_file("IOMMUFD device", constants::IOMMU_DEVICE_PATH)?;
+        } else if let Some(gpu_bdf_env_var) = Self::GPU_BDF_ENV_VAR_NAME {
```
sardok, line 69: Shouldn't we checking absence of `Self::GPU_BDF_ENV_VAR_NAME` instead? something like 
`if env[GPU_BDF_ENV_VAR_NAME] is None and env[ENABLE_GPU_PASSTHROUGH_ENV_VAR] == true: return error`?

alex-fortanix: This is the expected env var name per platform.

I did it this way so that we would have the env var name for the log (logging `TDX_GPU_BDF` for TDX and `SNP_GPU_BDF` for SNP`) at Rui's request.

sardok: Ok, i read it wrong, i thought `gpu_bdf_env_var` is the env variable value for a moment.

vsock-proxy/parent/src/platform/qemu.rs
```diff
@@ -55,6 +63,13 @@ pub(super) trait QemuPlatform {
...
+            if env::var(constants::ENABLE_GPU_PASSTHROUGH_ENV_VAR).is_ok_and(|v| v == "true") {
+                return Err(format!(
+                    "GPU passthrough was enabled at conversion time but {} env var is not set",
```
sardok, line 72: As i understand from this error message, `ENABLE_GPU_PASSTHROUGH_ENV_VAR` needs to be passed from container-converter.

alex-fortanix: Yes, with the recent changes it is part of this https://github.com/fortanix/salmiac/blob/beta/tools/container-converter/src/image_builder/parent/snp.rs#L47

sardok: Yes but parent env variables are defined here: https://github.com/fortanix/salmiac/blob/d7b8e91f2c88939f78ecd64e452dfb52977200b5/tools/container-converter/src/image_builder/parent/qemu.rs#L137. So i think, if `enable_gpu_passthrough` is `true`, then we should add `ENABLE_GPU_PASSTHROUGH=true` to that env_vars list, if i am not missing anything.

alex-fortanix: Ah yep I had that previously and wrongly removed it. I've added it back.

docker/qemu/parent-base/Dockerfile
```diff
@@ -124,7 +124,8 @@ RUN apt-get update \
 COPY --from=qemu-builder /opt/qemu /opt/qemu
 ENV PATH="/opt/qemu/bin:${PATH}"
 
-COPY --from=parent-base dnsmasq.d /etc/dnsmasq.d
+# This is represented as a template so that the build does not fail when dnsmasq.d is not present
+COPY --from=parent-base dnsmasq.[d] /etc/dnsmasq.d
```
sardok: Here, we make `dnsmasq.d` optional i guess? did you have a chance to try running enclave without this file (or directory)?

alex-fortanix: I am not able to verify currently as the SNP env is in use, but I have verified the failure cases.

I did this because `dnsmasq.d` was removed from SNP in https://github.com/fortanix/salmiac/pull/95/changes

sardok (approved):

vsock-proxy/parent/src/platform/tdx.rs
```diff
@@ -3,7 +3,6 @@
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-use std::env;
```
rui-alm, line 6: This is necessary for the `cpu` function.

rui-alm (changes requested):

vsock-proxy/parent/src/platform/tdx.rs
alex-fortanix: Whoops, added this back

rui-alm (approved):

PR #97, opened 2026-06-23
[RTE-1066]: Update qemu machine parameters for tdx.

sardok:
Updates the qemu parameters for Tdx based on the findings from ubuntu 26 host experiments. The most important update is the one made to `-machine`. If those additional arguments in this PR are missing, tdx-measure tool's measurements mismatch.

### DIFF
--- a/ENV_VARS.md
+++ b/ENV_VARS.md
@@ -55,6 +55,8 @@ registered as a debug build in a test-only deployment account.
   passed to the nitro-cli run command.
 
 ##### AMD SEV-SNP enclaves settings variables
+- CPU_COUNT - Override the `-smp` param passed while running the enclave i.e.
+  passed to the qemu run command.
 - MEM_SIZE - Override the `-m` param & memory backend object passed while running
   the enclave i.e. passed to the qemu run command.
 - SNP_GPU_BDF - For use with EnclaveOS containers with GPU passthrough.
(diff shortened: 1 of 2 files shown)


ENV_VARS.md
```diff
@@ -55,8 +55,6 @@ registered as a debug build in a test-only deployment account.
   passed to the nitro-cli run command.
 
 ##### AMD SEV-SNP enclaves settings variables
-- CPU_COUNT - Override the `-smp` param passed while running the enclave i.e.
-  passed to the qemu run command.
```
sardok, line 59: Removed this env variable because it affects measurement of sev/snp. In longer term will verify this or replace with cli argument(s) of vsock-pproxy parent.

jethrogb: You're not removing any functionality in this PR. Is this just updating stale documentation?

n7377747 (approved):

pavankalyan0424 (approved):

ENV_VARS.md
sardok: Well, this is not stale. This env variables works correctly if passed to converted image at runtime but breaks measurements. I am making it invisible in this PR till we land a better fix.

jethrogb: Better just document it in an issue for now and leave the documentation as is, especially since this PR is unrelated.

PR #101, opened 2026-06-30
RTE-1000: Make the TDX API models plural and follow PascalCase

n7377747:


api-model/src/tdx.rs
```diff
@@ -15,16 +15,16 @@ use crate::{
 
 #[derive(Clone, PartialEq, Debug)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct TDXEnclaveConversionRequest {
+pub struct TdxEnclavesConversionRequest {
     #[cfg_attr(feature = "serde", serde(flatten))]
     pub request: ConversionRequest, // Existing model, Refer to more details section above
     #[cfg_attr(feature = "serde", serde(rename = "tdx_enclave_options"))]
```
sardok: Should be `serde(rename = "tdx_enclaves_options")`.

PR #102, opened 2026-06-30
[RTE-949] Add SNP info and build/convert/run steps to README

alex-fortanix:


pavankalyan0424:
@alex-fortanix can you point it out to beta. Also beta branch has new changes, so docker comds related parent-base might change.

README.md
```diff
@@ -16,62 +16,70 @@ Salmiac enhances Nitro Enclaves by enabling networking for external communicatio
...
-   # build parent-base-nitro image
-   docker build -t parent-base-nitro nitro/parent-base
+   # Build parent-base-nitro image
+   docker build --no-cache docker/nitro/parent-base -t parent-base-nitro
+   
+   # Build parent-base-qemu image
+   docker build --no-cache \
+      --build-context parent-base=docker/nitro/parent-base \
+      -t parent-base-qemu \
+      docker/qemu/parent-base
```
pavankalyan0424: I think this is not needed for nitro

README.md
```diff
@@ -93,21 +101,160 @@ This guide allows you to build salmiac from source and convert your docker appli
    }
    ```
 
-7. Make your application Nitro VM-capable by running container converter with the file from previous step.
+6. Make your application Nitro VM-capable by running container converter with the file from previous step.
    The converter by default pulls the input image and pushes the output image to remote repositories. These images are then cleaned up from the local docker cache. In our example, the output image push is disabled in the request json and to preserve the images in the docker cache, 'PRESERVE_IMAGES' environment variable is specified.
    ```bash
-      docker run --rm -e PARENT_IMAGE=parent-base-nitro -e ENCLAVE_IMAGE=enclave-base --name nitro-converter --user 0 --privileged -v /var/run/docker.sock:/var/run/docker.sock -e PRESERVE_IMAGES=input,result -v /tmp/req.json:/app/req.json nitro-converter --request-file /app/req.json
+   docker run --rm \
+      -e PARENT_IMAGE=parent-base-qemu \
```
pavankalyan0424: This should be parent-base-nitro only

README.md
```diff
@@ -93,21 +101,160 @@ This guide allows you to build salmiac from source and convert your docker appli
...
+   ```
+
+   Execute `lspci -nnk -d 10de:` again, and if all is correct, it will be shown that the `vfio-pci` driver is in use.
+   
+6. To run the container, execute the following command, setting `SNP_GPU_BDF` using the value obtained previously, and set `APPCONFIG_ID` if one has been created for this instance.
+
+   The `APPCONFIG_ID` is the runtime configuration hash of the workflow, which can be created through the CCM UI. It is only necessary for workflows, and not needed if only the application is to be run in the container.
+
+   The variables `ENCLAVEOS_DEBUG` and `RUST_LOG` help with debugging any issues, and can be excluded if more detailed logging is not required.
+
```
pavankalyan0424, line 237: nit: Some of the information here duplicates USAGE.md. I feel we can guide user to that document for this kind of thing? or May be getting rid of Usage guide, and move everything here?

README.md
alex-fortanix: Oh I added this because I saw this is built off the nitro parent base in `.github/workflows/build-docker-images.yml`, I can remove it

README.md
alex-fortanix: I kind of lean towards removing `USAGE.md`, what do you think?

pavankalyan0424: works for me :)

README.md
alex-fortanix: Alright, fixed that

README.md
alex-fortanix: Fixed

README.md
alex-fortanix: Alright I have removed `USAGE.md`

pavankalyan0424 (approved):

pavankalyan0424:
@alex-fortanix , README side of changes looks good. But, I checked the ticket again, and looks like we will need to update `ENV_VARS.md` too
Currently the file misses env vars like RW_STORAGE_SIZE, ENCLAVEOS_APPCONFIG_ID, DSM_ENDPOINT, DSM_APP_ID

pavankalyan0424 (approved):

PR #104, opened 2026-07-01
[RTE-1074]: Fix vsock-proxy host device checks.

sardok:
This PR fixes the faulty host device checks in Tdx platform. Also adds minor fix to catch early enclave exits.

### DIFF
--- a/vsock-proxy/parent/src/parent.rs
+++ b/vsock-proxy/parent/src/parent.rs
@@ -40,7 +40,6 @@ use crate::network::{
     setup_network_devices, PairedPcapDevice, PairedTapDevice,
 };
 use crate::packet_capture::start_pcap_loops;
-use crate::platform::GuestLaunchResult;
 use crate::ParentConsoleArguments;
 
 const INSTALLATION_DIR: &str = "/opt/fortanix/enclave-os";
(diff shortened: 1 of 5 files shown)


vsock-proxy/parent/src/parent.rs
```diff
@@ -98,12 +99,25 @@ pub(crate) async fn run(args: ParentConsoleArguments) -> Result<UserProgramExitS
...
     info!("Spawning enclave process.");
-    let guest_launch_result = crate::platform::launch_guest();
-    // todo: will be used in https://fortanix.atlassian.net/browse/SALM-300
-    let _enclave_process = guest_launch_result.enclave_process;
+    let GuestLaunchResult { enclave_process } = crate::platform::launch_guest();
+    let vsock_stream = create_vsock_stream(VSOCK_PARENT_PORT);
 
     info!("Awaiting confirmation from enclave.");
-    let mut enclave_port = create_vsock_stream(VSOCK_PARENT_PORT).await?;
+    let mut enclave_port = tokio::select! {
```
alex-fortanix: I've got an open PR where I make this same change https://github.com/fortanix/salmiac/pull/96 Up to you, you can either remove this and hold until I get my PR in or I can remove it from mine and you can get it in here.

sardok: Ok, i have reverted mine.

rui-alm (approved):
Are we checking for the host vsock device? If not, maybe we should create a ticket to improve the checking of all required host devices.

sardok:
> Are we checking for the host vsock device? If not, maybe we should create a ticket to improve the checking of all required host devices.

Good catch, just pushed 404baf640e2282a014490241c7158ac02d9dc070.

PR #106, opened 2026-07-01
[RTE-1075] Allow passing CA list and CRL while authenticating with DSM

aditijannu:
This PR updates the `enclaveos-encrypted-fs` crate:
- Allows caller to pass a list of CA certs and CRL when creating a client to connect to DSM
- CA cert is optional only in debug builds to provide dev testing flexibility
- Includes unit tests to ensure DSM auth fails when an incorrect CA list is provided or when CA list is missing in release builds


tools/enclaveos-encrypted-fs/src/dsm_key_config.rs
```diff
@@ -567,6 +591,20 @@ mod tests {
         );
     }
 
+    fn load_ca_list() -> MbedtlsList<Certificate> {
+        let native = rustls_native_certs::load_native_certs();
+
+        let mut ca_list = MbedtlsList::<Certificate>::new();
+
+        for cert in native.certs {
+            if let Ok(parsed) = Certificate::from_der(cert.as_ref()) {
```
gilanghamidy, line 600: The `load_native_certs` are seemingly in PEM format, not in DER. So this will always fail, no?

https://docs.rs/rustls-native-certs/latest/rustls_native_certs/

tools/enclaveos-encrypted-fs/src/dsm_key_config.rs
```diff
@@ -567,6 +591,20 @@ mod tests {
...
+                ca_list.push(parsed);
+            }
+        }
+
+        ca_list
+    }
```
gilanghamidy, line 606: Probably it is good to have a unit test that validates this `load_ca_list` function perform correctly

gilanghamidy (commented):
So if I recall yesterday, the root certificate itself is already packaged in the converter, right? I assume that one is also in PEM format, not DER

tools/enclaveos-encrypted-fs/src/dsm_key_config.rs
gilanghamidy: Ah wait. The function itself returns `CertificateDer`. It is the certificate in the disk that is expected to be PEM. So I suppose this is correct then

tools/enclaveos-encrypted-fs/src/dsm_key_config.rs
aditijannu: `test_connection_to_dsm_with_appcert` test indirectly ensures that `load_ca_list` functions correctly. Otherwise the test would fail.

gilanghamidy (approved):

PR #107, opened 2026-07-02
[RTE-1008] TDX Converter Container

gilanghamidy:
Some refactoring for TDX converter container image. There's still some duplication in the `tdx` directory that I am still figuring out how to minimize that. Unfortunately, our internal build script cannot pass `--build-context` yet, so we cannot do similar job as in the `parent-base` container build that is handled entirely inside this repo.

Also moving all expected kernel artifact under the `staging` directory which is part of `.gitignore` already. Previously it was in the root directory of the `Dockerfile` which clobber the git tree.


docker/tdx/build-conv-container.sh
```diff
@@ -0,0 +1,18 @@
...
+cp ../../tools/container-converter/target/$1/container-converter ./staging
+mv ./staging/container-converter ./staging/server
+
+# Package the enclave and parent base images for the converter
+docker save -o ./staging/enclave-base.tar enclave-base
+docker save -o ./staging/enclave-base-gpu.tar enclave-base-gpu
+docker save -o ./staging/parent-base.tar parent-base-tdx
+
+# Build the converter
+docker build --no-cache -t snp-converter .
```
n7377747: `tdx-converter` ?

docker/tdx/Dockerfile
```diff
@@ -0,0 +1,64 @@
+# Download OVMF and EFI boot stubb
+ARG NVIDIA_DRIVER_BRANCH=595
+
+FROM ubuntu:25.10 AS apt_extract
```
n7377747, line 4: if we don't have any changes, can we use a common Dockerfile controlled by `PLATFORM` ARG?

n7377747: ignore, i see the base image changes.

docker/tdx/build-conv-container.sh
gilanghamidy: Ah yeah true. We really need to reduce this code duplication to streamline the source

docker/tdx/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+# These need to be pulled in from Jenkins https://jenkins.fortanix.net/job/build-salmiac-kernel/80/artifact/
+
+# Copy gpu disabled artifacts
+COPY staging/kernel_disabled_gpu/init /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/init
+COPY staging/tdx/kernel_disabled_gpu/.config /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/bzImage.config
+COPY staging/tdx/kernel_disabled_gpu/bzImage /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/bzImage
+
+# Copy gpu enabled artifacts
+COPY staging/kernel_enabled_gpu/init /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/init
```
n7377747, line 53: should this be `staging/tdx/kernel_enabled_gpu/` instead of `staging/kernel_enabled_gpu/`. 
I was thinking why not have tdx subfolder removed in this path since there're going to be no snp artifacts in this path.

docker/tdx/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+WORKDIR /app
+
+COPY --from=apt_extract /blob /opt/fortanix/enclave-os/blobs
+COPY --from=nvidia_userspace /nvidia-driver-payload /opt/fortanix/enclave-os/nvidia-driver-payload
+
+# These need to be pulled in from Jenkins https://jenkins.fortanix.net/job/build-salmiac-kernel/80/artifact/
+
+# Copy gpu disabled artifacts
+COPY staging/kernel_disabled_gpu/init /opt/fortanix/enclave-os/blobs/kernel_disabled_gpu/init
```
n7377747, line 48: same here as below

docker/snp/Dockerfile
```diff
@@ -61,18 +61,18 @@ COPY --from=nvidia_userspace /nvidia-driver-payload /opt/fortanix/enclave-os/nvi
...
-COPY kernel_enabled_gpu/.config /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/bzImage.config
-COPY kernel_enabled_gpu/bzImage /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/bzImage
-COPY kernel_enabled_gpu/nvidia-uvm.ko /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/nvidia-uvm.ko
-COPY kernel_enabled_gpu/nvidia.ko /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/nvidia.ko
-COPY kernel_enabled_gpu/nvidia-drm.ko /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/nvidia-drm.ko
-COPY kernel_enabled_gpu/nvidia-modeset.ko /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/nvidia-modeset.ko
+COPY staging/kernel_enabled_gpu/init /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/init
+COPY staging/snp/kernel_enabled_gpu/.config /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/bzImage.config
+COPY staging/snp/kernel_enabled_gpu/bzImage /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/bzImage
+COPY staging/snp/kernel_enabled_gpu/nvidia-uvm.ko /opt/fortanix/enclave-os/blobs/kernel_enabled_gpu/nvidia-uvm.ko
```
n7377747, line 72: same as below for snp, do we really need snp subfolder here?

docker/tdx/Dockerfile
gilanghamidy: Unfortunately this is because the internal CI script and Docker limitation that requires to copy everything into the directory under the Dockerfile location. The function in that internal script do batch copying of additional artifact into the `staging` directory.

Since @rui-alm has modified the directory structure of our kernel build, this is the simplest thing that we copy the entire `tdx` and `snp` directory instead of enumerating each files under these directory just to put it in the parent `staging` directory.

n7377747: ok

docker/tdx/stage-nvidia-driver-payload.sh
```diff
@@ -0,0 +1,156 @@
...
+  local src
+  local dst
+
+  for src in /lib/firmware/nvidia/*/gsp*.bin; do
+    [ -e "${src}" ] || continue
+
+    found=1
+    dst="${OUT_DIR}/firmware/${src#/}"
+
+    echo "Staging NVIDIA firmware: ${src} -> ${dst}"
```
n7377747, line 87: nit: we could either replace this with dst from src or make the other two functions follow src-> dst format for uniformity.

n7377747 (approved):

docker/tdx/stage-nvidia-driver-payload.sh
gilanghamidy: Yeah at the moment, it is copied and pasted from the SEV directory. I'll create an improvement ticket to actually refactor all these Dockerfile and bash scripting to simplify the image creation process

gilanghamidy:
I guess we can merge as is now, and @n7377747 can retarget your PR into the beta from here.

PR #108, opened 2026-07-02
[RTE-1075] Allow passing CA list and CRL while authenticating with DSM

aditijannu:
This PR updates the `enclaveos-encrypted-fs` crate:

- Allows caller to pass a list of CA certs and CRL when creating a client to connect to DSM
- CA cert is optional only in debug builds to provide dev testing flexibility
- Includes unit tests to ensure DSM auth fails when an incorrect CA list is provided or when CA list is missing in release builds

Corresponding PR that targets master branch - https://github.com/fortanix/salmiac/pull/106


gilanghamidy (approved):
Approve as reviewed with the other PR that targets master: https://github.com/fortanix/salmiac/pull/106

PR #110, opened 2026-07-06
Use tdxguest-acpi-tables tool when building acpi tables.

sardok:
Demonstrates how to use tdxguest-acpi-tables tool when acpi tables is needed for tdx measurements.


sardok:
`tdxguest-acpi-tables` tool is still highly unstable.

PR #112, opened 2026-07-07
RTE-998: TDX Converter container changes to support tdx-measure

n7377747:


tools/container-converter/src/image_builder/enclave/tdx_measurement.rs
```diff
@@ -0,0 +1,126 @@
...
+
+pub(crate) async fn compute_tdx_launch_measurement(
+    inputs: &TdxMeasurementInputs<'_>,
+) -> Result<TdxEnclavesMeasurements> {
+    let ovmf = inputs.ovmf.display().to_string();
+    let kernel = inputs.kernel.display().to_string();
+
+    let mut vm_objects = Vec::new();
+    // "reserve=off" to prevent qemu from reserving the memory during conversion
+    let memory_backend_device = format!(
```
sardok, line 74: IMO, this device shouldn't affect measurement.

tools/container-converter/src/image_builder/enclave/tdx_measurement.rs
```diff
@@ -0,0 +1,126 @@
...
+        "memory-backend-ram,id={},size={}M,reserve=off",
+        MEMORY_BACKEND_ID,
+        inputs.memory.to_mb()
+    );
+    vm_objects.push(memory_backend_device);
+    let mut vm_devices = Vec::new();
+    // TODO: Pending device addtion for gpu enabled devices
```
sardok: you can remove this `TODO:`

tools/container-converter/src/image_builder/enclave/tdx_measurement.rs
```diff
@@ -0,0 +1,126 @@
...
+    vm_devices.push(VSOCK_DEVICE.to_string());
+    let machine_arg = format!(
+        "q35,kernel_irqchip=split,memory-backend={MEMORY_BACKEND_ID},hpet=off,smm=off,pic=off",
```
sardok, line 85: If you can verify that above memory device is not necessary, you can remove `memory-backend` in this string.

tools/container-converter/src/image_builder/enclave/tdx_measurement.rs
```diff
@@ -0,0 +1,126 @@
...
+        )
+        .metadata_path(Path::new(""))
+        .patch_kernel(false)
+        .path_boot_xxxx("")
+        .boot_order("")
+        .table_loader("")
+        .rsdp("")
+        .build();
+
+    let measurements = machine.measure().expect("should not fail");
```
sardok: `.expect()` should be replaced by `?` (probably error conversion needed).

vsock-proxy/parent/src/platform/tdx.rs
```diff
@@ -55,6 +56,16 @@ impl QemuPlatform for TdxPlatform {
     fn gpu(&self) -> Option<String> {
         env::var("TDX_GPU_BDF").ok()
     }
+
+    fn globals(&self) -> Vec<String> {
+        let mut globals = vec![];
+
+        let gpu_device = self.gpu().as_ref().map(Self::build_vfio_iommu_arg);
+        if let Some(_) = &gpu_device {
```
sardok: Nit; you can don't need to map to `Self::build_vfio_iommu_arg`. `if self.gpu().is_some() { ...` should be enough.

vsock-proxy/parent/src/platform/tdx.rs
```diff
@@ -55,6 +56,16 @@ impl QemuPlatform for TdxPlatform {
...
+            globals.push(PCI_HOLE_SIZE_GLOBAL.to_string());
```
sardok, line 66: So without this, measurements are not matching i presume? For some reason i thought you wouldn't need this due to `fw_cfg`.

tools/container-converter/src/image_builder/enclave/tdx_measurement.rs
n7377747: but we need `reserve=off` to prevent upfront allocation

vsock-proxy/parent/src/platform/tdx.rs
n7377747: we need this, but we don't need `fw_cfg` for measurements

tools/container-converter/src/image_builder/enclave/tdx_measurement.rs
n7377747: we need it.

tools/container-converter/src/image_builder/enclave/tdx_measurement.rs
n7377747: this is the way to achieve it.

tools/container-converter/src/image_builder/enclave/tdx_measurement.rs
```diff
@@ -0,0 +1,162 @@
...
+        })
+        .map_err(|error| ConverterError {
+            message: format!(
+                "Failed to create temporary folder to hold OVMF for TDX build measurements : {}",
+                error
+            ),
+            kind: crate::ConverterErrorKind::EnclaveImageCreation,
+        })?;
+
+    let ovmf_path = temp_ovmf_dir.path().join("OVMF.fd");
```
gilanghamidy, line 59: Is this copying to temporary part necessary? Why can't we use the file directly?

n7377747: tdx-measure launches a container where it maps the ovmf path to another container, the tdx-measure container though launched from inside the converter-container, it cannot share any directory from it as it is launched from docker on host. so it only understands host directory and hence we need to create a temp file on host (we also do `-v /tmp:/tmp`) while launching converter so the ovmf can be shared with tdx-measure container. I have also added an env variable to override this directory.

sardok (approved):

PR #115, opened 2026-07-10
[RTE-1089] Fixing base images loading flow

gilanghamidy:
I noticed that the base image loading in the code is quite convoluted. The current logic is:
1. If environment variable is set, try loading from environment variable first from the host Docker repo
2. If it is not set, use hard-coded image name to try loading it from the host Docker repo
3. If it is still fails, load the tarball, then load again using the hard-coded image name

This creates an issue:
1. If image loading from the environment variable fails, it ignores the error and silently load from the second and third step, which is bad for user as it creates surprises.
2. If the host repo already has an image with the hard-coded name (especially if it pulls `latest` tag), it will load whatever images that already exists in the host repo, eventhough it may have been an outdated image compared to the shipped converter. Again, it creates surprises.
3. If the packaged tarball has different image name and tag, it will again fails when finishing step 3 because the Docker cannot find the hard-coded image name. It happens especially if we have version number as tag in the base images, while the Docker try to load `latest` which may not exist unless user manually or our loader programmatically adds the tag.

Therefore, I propose a change in the flow when loading the base images, which for both parent and enclave base images:
1. If environment variable `PARENT_BASE` or `ENCLAVE_BASE` is set, load from the environment variable and immediately fail if there is no such image in the repo
2. If the environment variable is not set, always load from the tarball, and pick up the image name and tag directly from the tarball. It will also overwrite the existing image with the same name and tag if it is already present but different hashes (Docker does this automatically)

Noteworthy remarks:
1. We can remove hard-coding base image name within the converter which reduces synchronization effort between different platforms and build scripts.
2. Function to load image from `liftship` crate (`images().load()`) unfortunately does not results in the actual loaded image, but only the stream of logs from the Docker daemon. We need to parse the output and find the loaded image name from the log stream.
3. `liftship` crate `Reference` instance (aliased as `DockerReference` in our code) uses a borrowed reference to the input string, which results in some reference trickery in the code to allow returning the `DockerReference` from `load_base_image` function correctly to its caller.

### DIFF
--- a/tools/container-converter/src/docker.rs
+++ b/tools/container-converter/src/docker.rs
@@ -232,7 +232,7 @@ impl DockerUtil for DockerDaemon {
                     "Failed pulling image {} from remote repository. {:?}",
                     image.to_string(),
                     err
-                ))
+                ));
             }
             Ok(_) => {}
         }
@@ -264,7 +264,16 @@ impl DockerUtil for DockerDaemon {
                     match output {
                         ImageBuildChunk::Update { stream } => {
                             if let Some(image_name) = stream.strip_prefix("Loaded image: ") {
-                                loaded_image_name = Some(image_name.trim().to_owned());
+                                if loaded_image_name.is_none() {
+                                    loaded_image_name = Some(image_name.trim().to_owned());
+                                } else {
+                                    return Err(format!(
+                                        "Loaded docker image {:?} contains more than one valid images. Previously loaded image was {:?}, while another image is {:?}",
+                                        tar_path,
+                                        loaded_image_name.unwrap(),
+                                        image_name
+                                    ));
+                                }
                             }
                         }
                         _ => (),
--- a/tools/container-converter/src/docker.rs
+++ b/tools/container-converter/src/docker.rs
@@ -15,7 +15,7 @@ use docker_image_reference::{Reference as DockerReference, Reference};
 use futures::StreamExt;
 use log::{debug, error, info, warn};
 use shiplift::container::ContainerCreateInfo;
-use shiplift::image::{BuildParams, ImageDetails, PushOptions};
+use shiplift::image::{BuildParams, ImageBuildChunk, ImageDetails, PushOptions};
 use shiplift::{
     ContainerOptions, Docker, Image, PullOptions, RegistryAuth, RmContainerOptions, TagOptions,
 };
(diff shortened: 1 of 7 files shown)


tools/container-converter/src/lib.rs
```diff
@@ -402,68 +423,108 @@ fn hex_response(arg: &str) -> Result<HexString> {
...
-                .map_err(|message| ConverterError {
-                    message: format!("Failed retrieving requisite {} image. {:?}", image, message),
+                })?,
+        );
+
+        let loaded_image_name = image.as_ref().unwrap();
+
+        info!(
+            "Loaded requisite from backup tar file: {}",
+            loaded_image_name
```
sardok: I guess this should be `base_image_tar_path`?

tools/container-converter/src/lib.rs
```diff
@@ -172,10 +172,22 @@ async fn run0(
 ) -> Result<PlatformConversionResponse> {
     validate_request(&conversion_request.request)?;
 
-    let parent_image = env::var("PARENT_IMAGE").unwrap_or(PARENT_IMAGE.to_string());
```
sardok, line 176: Now some variables `PARENT_IMAGE`, `ENCLAVE_IMAGE` & `ENCLAVE_IMAGE_*_GPU` became unused. Better be removed.

tools/container-converter/src/docker.rs
```diff
@@ -247,17 +247,28 @@ impl DockerUtil for DockerDaemon {
...
+
         while let Some(import_result) = stream.next().await {
             match import_result {
                 Ok(output) => {
-                    info!("{:?}", output)
+                    info!("{:?}", output);
+
+                    match output {
+                        ImageBuildChunk::Update { stream } => {
+                            if let Some(image_name) = stream.strip_prefix("Loaded image: ") {
```
sardok, line 266: A tar file can contain multiple images (i.e `docker save -o output.tar img1:latest img2:latest img3:latest`). It seems to be unlikely in our case but i think we should give a warning (or error?) if `loaded_image_name` is already set & being overriden.

gilanghamidy: Ah this is one I did not know before. I will check this behavior and have proper guards to refuse this situation.

tools/container-converter/src/lib.rs
gilanghamidy: I am rewording the string to include both information

tools/container-converter/src/docker.rs
aditijannu: Don't we need to package both enclave-base-gpu and enclave-base? Or I want to understand how we pick the enclave base image when gpu passthrough is enabled/disabled?

gilanghamidy: In the cotainer converter container, we are packaging both non-GPU and GPU enclave base images. The non GPU is shared with the Nitro one, while the GPU one is specific to either TDX or SEV, and will be packaged according to the specific platform.

aditijannu (approved):

PR #116, opened 2026-07-15
RTE-1095 - Disable ACPI table creation and exclude them from rtmr0

rui-alm:
Depends on https://github.com/fortanix/tdx-measure/pull/4


sardok:
I think we woould need `Cargo.toml` & `Cargo.lock` file changes in this PR once tdx-measure PR is merged.

sardok (approved):

PR #118, opened 2026-07-16
[RTE-1090] Use Custom OVMF in converted container

gilanghamidy:
Tested separately in the internal CI and it is working. The OVMF itself is already included when the CI is preparing the `staging` content in the `docker` directory.

The existing code to pull OVMF from the online repo is commented out, but I propose to keep it there in case we want to reverting to using mainline OVMF.

### DIFF
--- a/docker/tdx/Dockerfile
+++ b/docker/tdx/Dockerfile
@@ -1,4 +1,4 @@
-# Download OVMF and EFI boot stubb
+# Download OVMF and EFI boot stub
 ARG NVIDIA_DRIVER_BRANCH=595
 
 FROM ubuntu:25.10 AS apt_extract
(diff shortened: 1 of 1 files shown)


docker/tdx/Dockerfile
```diff
@@ -10,9 +10,9 @@ RUN apt-get download systemd-boot-efi && \
     dpkg -x systemd-boot-efi*.deb system_boot_efi_dir && \
     cp system_boot_efi_dir/usr/lib/systemd/boot/efi/linuxx64.efi.stub /blob
```
rui-alm, line 11: Is this used anywhere?

rui-alm (approved):

docker/tdx/Dockerfile
gilanghamidy: I am not entirely sure but since it is copied alongside the `/blob` directory, I assume it was used somewhere. But you know more :)

rui-alm: I'm pretty sure it isn't, but we can tidy it up later.

PR #119, opened 2026-07-16
[RTE-1102] Fixing build issues in `beta` Branch

gilanghamidy:
Previous merge was slightly broken and did not pass `sanity.sh` and some unit tests. This PR fixes that.

I am changing the way we perform unit testing of the QEMU args generator, so now it can catch the issue accordingly if the expected and actual arguments differ. I identified new arguments being introduced, which are: `-nodefaults` and `-serial`.

### DIFF
--- a/tools/container-converter/src/image_builder/enclave/mod.rs
+++ b/tools/container-converter/src/image_builder/enclave/mod.rs
@@ -49,9 +49,9 @@ use api_model::converter::{CertificateConfig, ConverterOptions, DsmConfiguration
 use api_model::enclave::EnclaveManifest;
 use api_model::enclave::{CcmBackendUrl, FileSystemConfig, UserConfig};
 #[cfg(any(platform = "snp", platform = "tdx"))]
-use api_model::{ByteUnit, EnclavesOptions};
-#[cfg(any(platform = "snp", platform = "tdx"))]
 use api_model::NvidiaDriverCapability;
+#[cfg(any(platform = "snp", platform = "tdx"))]
+use api_model::{ByteUnit, EnclavesOptions};
 use docker_image_reference::Reference as DockerReference;
 use log::{debug, info, warn};
 use nix::sys::statfs::statfs;
@@ -176,7 +176,6 @@ impl EnclaveSettings {
             cpu_count: enclaves_options.cpu_count,
             #[cfg(any(platform = "snp", platform = "tdx"))]
             mem_size: enclaves_options.mem_size.clone(),
-
         }
     }
 }
@@ -897,7 +896,7 @@ impl<'a> GenericEnclaveImageBuilder<'a> {
                 return Err(ConverterError {
                     message: format!("Failed stat for {}. {:?}", path.display(), err),
                     kind: ConverterErrorKind::BlockFileCreation,
-                })
+                });
             }
         }
 
--- a/vsock-proxy/parent/src/platform/tdx.rs
+++ b/vsock-proxy/parent/src/platform/tdx.rs
@@ -89,9 +89,9 @@ async fn start_tdx_guest() -> Result<(), String> {
 
 #[cfg(test)]
 mod tests {
-    use std::collections::HashSet;
-
     use super::*;
+    use crate::platform::qemu::tests::diff_args;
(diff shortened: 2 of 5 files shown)


tools/container-converter/src/image_builder/enclave/mod.rs
```diff
@@ -49,9 +49,9 @@ use api_model::converter::{CertificateConfig, ConverterOptions, DsmConfiguration
 use api_model::enclave::EnclaveManifest;
 use api_model::enclave::{CcmBackendUrl, FileSystemConfig, UserConfig};
 #[cfg(any(platform = "snp", platform = "tdx"))]
-use api_model::{ByteUnit, EnclavesOptions};
-#[cfg(any(platform = "snp", platform = "tdx"))]
 use api_model::NvidiaDriverCapability;
```
n7377747, line 52: is this change a mistake by any chance?

n7377747: It doesn't appear to be a refactoring change.

gilanghamidy: It rejects the `cargo fmt --check`. This fixes that. It just changes the ordering in which it appears, as it is replaced below

n7377747: i didn't notice the feature flag above the NvidiaDriverCapability

n7377747 (approved):

vsock-proxy/parent/src/platform/tdx.rs
```diff
@@ -108,11 +108,11 @@ mod tests {
...
+            "-serial", "mon:stdio",
+            "-nodefaults",
         ];
-        let expected: Vec<String> = expected.into_iter().map(String::from).collect();
-        let expected: HashSet<String> = expected.into_iter().collect();
         let platform = TdxPlatform {};
-        let args: HashSet<String> = platform.build_qemu_args().into_iter().collect();
-        assert_eq!(expected, args, "QEMU args mismatched",);
+        let args: Vec<String> = platform.build_qemu_args().into_iter().collect();
+        diff_args(&expected, &Vec::from_iter(args.iter().map(String::as_str)));
```
rui-alm, line 116: It looks like the previous code has some issues that I'm afraid are still present.
The diff args seems to be converting the vectors of expected and actual qemu arguments, by converting them to a `HashMap` (key value pairs), ensuring the intersection matches and any elements present/missing on one but not the other are also detected.

It also seems to be doing a validation of QEMU parameters by ensuring no value can exist without a key, such as, `4G -m`. However, although it may be valid in our current case, I don't think it's valid as per QEMU usage:
`usage: qemu-system-x86_64 [options] [disk_image]`

A disk image can be provided without a key.

More importantly, I believe the `diff_args` (as the previous code with `HashSet`) will miss duplicate keys, for example, passing `-m 4G` and `-m 8G`, which, regardless of whether QEMU accepts it or not, there would be no reason for our code to produce such arguments.

Although comparing (orderly) the content of 2 vectors using `assert_eq!` isn't exactly how unit tests should work, since the order of the arguments won't actually matter in our case (but would in the case of a disk image), I think I still prefer it over the key value pair or HashSet approach. Also, `assert_eq!` over two `HashMap` should fail if any entries are missing.

gilanghamidy: > It also seems to be doing a validation of QEMU parameters by ensuring no value can exist without a key, such as, 4G -m

In this case, the `diff_args` will reject automatically since there's no leading argument key (which has `-` character). And besides, I was assuming that we only validates `[options]` part since that is the part our codes generate. So, the `[disk_image]` value should never appear in both expected and actual.

This unit test is at least allowing us to easily capture new arguments being introduced in the parent `Qemu` struct, or `Tdx` and `Sev` structs without updating the unit test. We don't need to cover the entirety that it is being accepted to QEMU command line, since in that case, we may need to implement the actual QEMU argument parser in our unit test code, which is unnecessary.

gilanghamidy:
Merging as it is now, further improvement in unit testing cases will be done in the next dev iteration.

PR #122, opened 2026-07-20
[RTE-1109] Dockerfiles Hardening

gilanghamidy:
Add mising `apt-get upgrade` and `apt-get clean` in Dockerfile.

As discussed with @rui-alm, doing `apt-get update`/`upgrade` in the enclave base image may indeed also upgrade the NVIDIA driver library. We just need to make sure that the kernel driver we compile is synced with the userspace library we package. We may prefer to try pinning the specific minor and revision version as well in our Dockerfile argument, but I think it will cause more confusion and complexity for now.

According to Rui, the risk at the moment is that there is a race condition between the kernel compilation and the Salmiac converter compilation, in which if Canonical publishes a version bump to the NVIDIA drivers in between, the converted image may not work due to the older kernel with newer userspace library. But I think this will be an unlikely situation for now.

### DIFF
--- a/docker/nitro/parent-base/Dockerfile
+++ b/docker/nitro/parent-base/Dockerfile
@@ -62,7 +62,9 @@ RUN apt-get update \
         systemctl \
         strace \
         tcpdump \
-    && apt-get upgrade -y
+    && apt-get upgrade -y \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists
 
 # Copy nitro-cli files
 COPY --from=nitro-cli aws-nitro-enclaves-cli/install/etc/nitro_enclaves /etc/nitro_enclaves
--- a/docker/qemu/parent-base/Dockerfile
+++ b/docker/qemu/parent-base/Dockerfile
@@ -14,6 +14,7 @@ ARG QEMU_TARBALL_SHA256
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update \
+    && apt-get upgrade -y \
     && apt-get install -y --no-install-recommends \
         bison \
         build-essential \
@@ -33,6 +34,7 @@ RUN apt-get update \
         python3 \
         python3-venv \
         xz-utils \
+    && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
 # Download the official QEMU release tarball, verify its SHA-256, then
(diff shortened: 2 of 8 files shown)


docker/simulator/Dockerfile
```diff
@@ -6,8 +6,9 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         ca-certificates \
         cryptsetup-bin \
-        sudo \
-    && rm -rf /var/lib/apt/lists/*
+        sudo && \
+    apt-get clean && \
```
rui-alm, line 11: Should we do upgrade on the simulator as well?

docker/qemu/parent-base/Dockerfile
rui-alm, line 16: Should we add upgrade here as well? It's not as relevant as the build tools installation should pull in up-to-date dependencies, and the parent base will do the upgrade, but at least we could claim the QEMU build happened in a fully up to date container, just in case there's some chance of exploiting a vulnerability on a different component not covered by the dependencies below.

docker/snp/Dockerfile
rui-alm: For the same reasons as QEMU, should we upgrade here as well to ensure sev snp measurement is built in a fully up to date container?

docker/qemu/parent-base/Dockerfile
gilanghamidy: Wait, I thought I also put upgrade down below here....

rui-alm: The `parent-base-qemu` already has the upgrade, which means the final parent-base image will have all of its packages up to date, but this section here, part of `qemu-builder` doesn't.
This is a temporary section used to build QEMU before we copy the built artifacts to the `parent-base-qemu`. My suggestion was to do upgrade here as well, so we know the build happened on a fully up to date environment.

docker/simulator/Dockerfile
gilanghamidy: This was missing indeed. Fixed

rui-alm (approved):

docker/nitro/parent-base/Dockerfile
```diff
@@ -62,7 +62,9 @@ RUN apt-get update \
         systemctl \
         strace \
         tcpdump \
-    && apt-get upgrade -y
+    && apt-get upgrade -y \
```
aditijannu, line 65: The system needs to be upgraded before installing packages. The packages installed from line 51 will still be outdated. Same comment for all the other docker files.

PR #134, opened 2026-08-23
[RTE-1153] Enable TDX build in github actions

pavankalyan0424:


build-solution.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -exo pipefail
+
+SALMIAC_PLATFORMS=(
+    nitro
+    snp
+    tdx
```
sardok, line 8: Can you add `simulator` mode as well?

sardok (approved):

PR #135, opened 2026-08-23
[RTE-1150] Fix Converter API documentation

pavankalyan0424:
This PR fixes Converter API documentation

### DIFF
--- a/api-model/docs/models.openapi.yaml
+++ b/api-model/docs/models.openapi.yaml
@@ -21,10 +21,10 @@ components:
     CaCertificateConfig:
       type: "object"
       properties:
-        caPath:
+        ca_path:
           type: "string"
           description: "Path to expose the CA cert in the application filesystem"
-        caCert:
+        ca_cert:
           type: "string"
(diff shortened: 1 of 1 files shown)


gilanghamidy:
Can you also remove the `req.json` file that you proposed to get removed in the other PR? I think this PR can cover that as well

pavankalyan0424:
> Can you also remove the `req.json` file that you proposed to get removed in the other PR? I think this PR can cover that as well

@gilanghamidy I rechecked this and I think we should us keep the file. I see that it is used by push-enclaveos-image.sh, push-converter-image.sh, run-container-converter-interactive.sh, run-container-converter.sh.

So I think we should keep the file.

sardok:
> I see that it is used by push-enclaveos-image.sh, push-converter-image.sh, run-container-converter-interactive.sh, run-container-converter.sh.

I think at some point we should move these scripts to a specific directory (`scripts` for example?) and unify them.

sardok (approved):

PR #136, opened 2026-08-24
[RTE-1141] Add multi GPU support.

sardok:
This PR adds multi-GPU support by introducing new env variable to be used `TDX_GPU_BDFS` & `SNP_GPU_BDFS`, while keeping the existing env variables (probably i'll remove them later).

### DIFF
--- a/docker/enclave-base-gpu/Dockerfile
+++ b/docker/enclave-base-gpu/Dockerfile
@@ -11,7 +11,13 @@ ARG NVIDIA_DRIVER_BRANCH=595
 
 FROM ubuntu:${UBUNTU_VERSION} AS enclave-base-gpu
 ARG NVIDIA_DRIVER_BRANCH
+ARG NVIDIA_DRIVER_BRANCH_PINNED=595.71.05-0ubuntu0.24.04.1
+ARG UBUNTU_SNAPSHOT=20260601T000000Z
 ENV DEBIAN_FRONTEND=noninteractive
+
+# Bootstrap HTTPS trust using the current archive first.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends ca-certificates
 
 # Runtime packages. Groups:
 #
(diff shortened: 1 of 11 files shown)


docker/enclave-base-gpu/Dockerfile
```diff
@@ -27,14 +31,17 @@ ENV DEBIAN_FRONTEND=noninteractive
...
+    && apt-get update \
     && apt-get install -y --no-install-recommends \
         ca-certificates \
         cryptsetup \
         nbd-client \
-        libnvidia-compute-${NVIDIA_DRIVER_BRANCH} \
-        nvidia-utils-${NVIDIA_DRIVER_BRANCH} \
+        libnvidia-compute-${NVIDIA_DRIVER_BRANCH}=${NVIDIA_DRIVER_BRANCH_PINNED} \
+        nvidia-utils-${NVIDIA_DRIVER_BRANCH}=${NVIDIA_DRIVER_BRANCH_PINNED} \
         openssl \
```
sardok, line 45: I tried installing particular firmware, our driver expects but none of them worked (always ends up driver/library incompatibility). The only thing that seems to working is freezing the ubuntu repositories at certain date.

vsock-proxy/parent/src/platform/qemu.rs
```diff
@@ -20,23 +23,77 @@ pub(super) mod constants {
...
+    pub bdfs: Vec<String>,
+    pub mmio64_mb: u64,
+}
+
+impl GPUSettings {
+    pub fn try_new_from_env_variables(
+        plural_env_var: Option<&'static str>,
+        singular_env_var: Option<&'static str>,
+    ) -> Result<Option<Self>, String> {
+        let plural = plural_env_var.and_then(|name| env::var(name).ok());
```
gilanghamidy: Is it necessary to split the environment variable into two? Can we just use a single environment variable and parse each element? Meaning that if it only has a single entry, then it is a single pass through, while if it is multiple entries, it is multiple GPUs.

sardok: I did that to make it backward compatible then, started planning to drop it in the following PR. Maybe i should  do it with this PR.

rui-alm: If we use `GPU_BDF=0001;0002` and then split on `;`, we'd still be backwards compatible, right?

sardok: Yes, i have been splitting by `,`, should it be `;`? The BDF addresses contain `:` and `.` if i'm not wrong, so both `,` & `;` should be ok.

About the env variables, which makes the most sense to you? `GPU_BDFS`, `GPU_BDF` (note that current env vars are `TDX_GPU_BDF` & `SNP_GPU_BDF`)?

@rui-alm @gilanghamidy

rui-alm: Commas should be fine. I was mentioning that for the old variables `TDX_GPU_BDF` and `SNP_GPU_BDF`, we'd still be backwards compatible if we had multiple values with a separator, if we'd want to support it. It could be a bit odd, since BDF is still singular, but nothing too surprising. There's env vars that are singular, like `PATH`, and it takes multiple values.

For a new env var, `GPU_BDFS` (plural) makes more sense to me, it implies multiple are supported, but you fill in how many you want, including a single one.

sardok: Ok, for the sake of backard compatibility, i kept `TDX_GPU_BDF` & `SNP_GPU_BDF` as they are. So those env variables may contain multiple bdf addresses now.

gilanghamidy: Sinan and I discussed yesterday, and these CLI environment variable warrants an API review. Since everything is still in `beta` at this stage, I believe we can make breaking changes and does not necessarily need to strictly adhere backwards compatibility. We should consult more with relevant people and wait for @aditijannu to return back from holiday next week about naming and standardization. However, we can just name it and merge it as-is for now, and as Sinan mentioned in the Slack, we can standardize all environment variable naming + API review process in a separate ticket.

rui-alm (approved):

PR #137, opened 2026-08-24
[RTE-965] Supporting multiple Salmiac enclave launch in a host

gilanghamidy:
Following similar behavior to Fortanix VME Runner.

### DIFF
--- a/vsock-proxy/enclave/src/main.rs
+++ b/vsock-proxy/enclave/src/main.rs
@@ -22,10 +22,10 @@ use shared::NumArg;
 fn obtain_vsock_cid() -> Result<u32, String> {
     mod constants {
         pub const VSOCK_DEVICE_PATH: &str = "/dev/vsock";
-        pub const IOCTL_VM_SOCKETS_GET_LOCAL_CID: usize = 0x7b9;
+        // https://github.com/torvalds/linux/blob/786262be6048deab760f68c8acc2c85607165894/include/uapi/linux/vm_sockets.h#L196
+        pub const IOCTL_VM_SOCKETS_GET_LOCAL_CID: u64 = nix::request_code_none!(0x7, 0xb9);
     }
 
-    // linux/include/uapi/linux/vm_sockets.h
     nix::ioctl_read_bad!(
         get_local_cid,
         constants::IOCTL_VM_SOCKETS_GET_LOCAL_CID,
@@ -42,7 +42,7 @@ fn obtain_vsock_cid() -> Result<u32, String> {
 
     debug!("Obtained CID: {cid}");
 
-    Ok(cid as u32)
+    Ok(cid)
 }
 
 #[tokio::main(flavor = "multi_thread", worker_threads = 2)]
--- a/vsock-proxy/enclave/src/main.rs
+++ b/vsock-proxy/enclave/src/main.rs
@@ -16,7 +16,34 @@ use std::process;
 use clap::{App, AppSettings, Arg, ArgMatches};
 use log::{debug, error};
 use shared::models::UserProgramExitStatus;
-use shared::{parse_console_argument, NumArg};
+use shared::NumArg;
+
+#[cfg(any(platform = "snp", platform = "tdx", platform = "simulator"))]
+fn obtain_vsock_cid() -> Result<u32, String> {
+    mod constants {
+        pub const VSOCK_DEVICE_PATH: &str = "/dev/vsock";
+        pub const IOCTL_VM_SOCKETS_GET_LOCAL_CID: usize = 0x7b9;
+    }
+
+    // linux/include/uapi/linux/vm_sockets.h
+    nix::ioctl_read_bad!(
+        get_local_cid,
+        constants::IOCTL_VM_SOCKETS_GET_LOCAL_CID,
+        u32
+    );
+
+    use nix::{fcntl::OFlag, sys::stat::Mode};
+
+    let fd = nix::fcntl::open(constants::VSOCK_DEVICE_PATH, OFlag::O_RDONLY, Mode::empty())
+        .map_err(|e| format!("unable to open vsock device: {e}"))?;
+
+    let mut cid: u32 = 0;
+    unsafe { get_local_cid(fd, &mut cid).map_err(|e| format!("unable to get local cid: {e}"))? };
+
+    debug!("Obtained CID: {cid}");
+
+    Ok(cid as u32)
+}
 
 #[tokio::main(flavor = "multi_thread", worker_threads = 2)]
 async fn main() -> Result<(), String> {
@@ -24,7 +51,12 @@ async fn main() -> Result<(), String> {
 
     let matches = console_arguments();
 
-    let vsock_port = parse_console_argument::<u32>(&matches, "vsock-port");
+    #[cfg(platform = "nitro")]
+    let vsock_port = shared::VSOCK_PARENT_PORT;
+
+    #[cfg(any(platform = "snp", platform = "tdx", platform = "simulator"))]
+    let vsock_port = obtain_vsock_cid()?;
+
     let settings_path = matches
         .value_of("settings-path")
         .map(|e| Path::new(e))
--- a/vsock-proxy/parent/src/platform/mod.rs
+++ b/vsock-proxy/parent/src/platform/mod.rs
@@ -9,15 +9,14 @@ use tokio::task::JoinHandle;
 
 #[cfg(platform = "nitro")]
 mod nitro;
+#[cfg(any(platform = "snp", platform = "tdx", platform = "simulator"))]
+mod qemu;
 #[cfg(platform = "simulator")]
 mod simulator;
 #[cfg(platform = "snp")]
 mod snp;
 #[cfg(platform = "tdx")]
 mod tdx;
-
-#[cfg(any(platform = "snp", platform = "tdx", platform = "simulator"))]
-mod qemu;
 
 pub(crate) type GuestTasks = FuturesUnordered<JoinHandle<Result<(), String>>>;
 
@@ -35,6 +34,12 @@ pub(crate) use tdx::{launch_guest, should_forward_client_logs, start_post_connec
 #[cfg(platform = "nitro")]
 pub(crate) use nitro::{launch_guest, should_forward_client_logs, start_post_connect_guest_tasks};
 
+#[cfg(any(platform = "snp", platform = "tdx", platform = "simulator"))]
+pub(crate) use qemu::VmConnectionConfig;
+
 pub(crate) struct GuestLaunchResult {
     pub(crate) enclave_process: JoinHandle<Result<(), String>>,
+
+    #[cfg(any(platform = "snp", platform = "tdx", platform = "simulator"))]
+    pub(crate) enclave_connection_config: VmConnectionConfig,
 }
--- a/vsock-proxy/parent/src/platform/qemu.rs
+++ b/vsock-proxy/parent/src/platform/qemu.rs
@@ -3,11 +3,17 @@
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+use nix::fcntl::OFlag;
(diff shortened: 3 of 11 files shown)


vsock-proxy/parent/src/platform/tdx.rs
```diff
@@ -112,7 +113,11 @@ mod tests {
             "-nodefaults",
         ];
         let platform = TdxPlatform {};
-        let args: Vec<String> = platform.build_qemu_args().into_iter().collect();
+        let f = File::open("/dev/null").unwrap();
```
sardok: i don't think you need to open an actual device. `RawFd` is an alias to `int`. So you should be able to pass a constant/static number like `10` for example.

vsock-proxy/parent/src/platform/qemu.rs
```diff
@@ -99,7 +107,46 @@ pub(super) trait QemuPlatform {
         vec![]
     }
 
-    fn build_qemu_args(&self) -> Vec<String> {
+    /// Based on Fortanix VME runner implementation (amdsevsnp.rs: get_available_guest_cid_with_fd)
+    /// We may want to centralize the functionality elsewhere to avoid code duplication
+    fn get_available_guest_cid_with_fd() -> Result<(OwnedFd, u64), String> {
```
sardok: In my current open PR, i created `utils/qemu.rs`. See it here: https://github.com/fortanix/salmiac/pull/136/changes#diff-fa2698ce5c55c8a1a27563b768140fef9ae50830407857555983cbd0a6c9df0d. If you follow the same structure, then you may consider moving `get_available_guest_cid_with_fd`  & `set_guest_cid` to qemu utils file.

sardok: Or i can do it in my current PR too if this one gets merged before.

vsock-proxy/parent/src/platform/qemu.rs
```diff
@@ -153,6 +200,13 @@ pub(super) trait QemuPlatform {
             ]);
         }
 
+        let vsock_device_str = format!(
+            "vhost-vsock-pci,id=vhost-vsock-pci0,vhostfd={},guest-cid={}",
```
sardok, line 222: `id=vhost-vsock-pci0` is unnecessary here. IMO, we should consider adding `id` for a particular device is, if it is referred by other `-object` or `-device` (like `mem0`, `iommufd0`). I think it is coming from fortanix-vme, at that time we were kind of discovering qemu args and basically following the examples.

vsock-proxy/parent/src/platform/qemu.rs
```diff
@@ -99,7 +107,46 @@ pub(super) trait QemuPlatform {
...
+                            "unable to ioctl vhost-vsock device. Error code {err_code}"
+                        ));
+                    }
+                }
+            }
+        }
+        Err(format!("unable to get available CID for the guest"))
+    }
+
+    fn build_qemu_args(&self, vsock_fd: RawFd, vsock_cid: u64) -> Vec<String> {
```
sardok: When working on this abstract qemu platform, i tried to move all logically similar variables into trait functions with default implementations, like `QemuPlatform::firmware()`, `QemuPlatform::machine()`, `QemuPlatform::gpu()` etc, so we won't have to pass dozen of arguments to qemu string builder. 
I think, vhost configuration fits into this pattern too. I would suggest to add something like;

```
struct VsockConfig {
    vsock_fd: ..,
    vsock_cid: ...,
}

trait QemuPlatform {
    ...
    fn vsock_config(&self) -> Option<VsockConfig> { // Could be straight `VsockConfig` too, as we always expect to bind it?
          // Execute get_available_guest_cid_with_fd() etc
          ...
     }

     fn build_qemu_args() -> Vec<String> {
          ...
          if let Some(vsock_config) = self.vsock_config() {
                 let VsockConfig { vsock_fd, vsock_cid } = vsock_config;
                 let vsock_device_str = format!(
                      "vhost-vsock-pci,id=vhost-vsock-pci0,vhostfd={},guest-cid={}",
                      vsock_fd, vsock_cid
                 );
                args.extend(["-device", vsock_device_str.as_str()]);
          }

          ....
     }
     ...
}
```
What do you think?

gilanghamidy: This is actually quite nice, but I refactored the way we launch QEMU. Maybe you can take a look with my refactor in this PR. Because with my PR, determining CID + parent port has to happen before launching the VM. I centralize everything now in a single place and simplify some `async` function.

vsock-proxy/parent/src/platform/tdx.rs
gilanghamidy: YEah that's true lol. We only use this code to do validation of the QEMU args builder, not actually using a real FD

vsock-proxy/parent/src/platform/qemu.rs
sardok: @gilanghamidy i've just discovered that, explicit `fd` passing while leaving `id` empty is crashing Qemu for vfio device. I don't if this is same for vsock but keep this in mind, if you see a crash, can be related with this as well.

gilanghamidy: @sardok then maybe I should keep this `id` argument here then. In our `enclave-runner`, it is like this.

vsock-proxy/parent/src/platform/qemu.rs
gilanghamidy: I still put it inside the `platform/qemu.rs` as it has some data structure that is exposed to the parent. Should we just put all `qemu` related functions under the `qemu.rs` file itself? Since I don't think it will be used elsewhere

vsock-proxy/enclave/src/main.rs
```diff
@@ -16,15 +16,47 @@ use std::process;
...
 use log::{debug, error};
 use shared::models::UserProgramExitStatus;
-use shared::{parse_console_argument, NumArg};
+use shared::NumArg;
+
+#[cfg(any(platform = "snp", platform = "tdx", platform = "simulator"))]
+fn obtain_vsock_cid() -> Result<u32, String> {
+    mod constants {
+        pub const VSOCK_DEVICE_PATH: &str = "/dev/vsock";
+        pub const IOCTL_VM_SOCKETS_GET_LOCAL_CID: usize = 0x7b9;
```
sardok: Interesting, is this working? i would expect to see something:

```
const IOCTL_VM_SOCKETS_GET_LOCAL_CID: libc::c_ulong =
    nix::request_code_none!(7, 0xb9);
```

sardok: Ok, that macro result to the same value.

vsock-proxy/enclave/src/main.rs
```diff
@@ -16,15 +16,47 @@ use std::process;
...
+    }
+
+    // linux/include/uapi/linux/vm_sockets.h
```
sardok: Nit, permalink just in case https://github.com/torvalds/linux/blob/1b78070aaef63512688aebfbc82365ef9d6660f1/include/uapi/linux/vm_sockets.h#L196

vsock-proxy/enclave/src/main.rs
```diff
@@ -16,15 +16,47 @@ use std::process;
...
     env_logger::init();
 
     let matches = console_arguments();
 
-    let vsock_port = parse_console_argument::<u32>(&matches, "vsock-port");
+    #[cfg(platform = "nitro")]
+    let vsock_port = shared::VSOCK_PARENT_PORT;
+
+    #[cfg(any(platform = "snp", platform = "tdx", platform = "simulator"))]
+    let vsock_port = obtain_vsock_cid()?;
```
sardok, line 58: Hmm, i am little bit confused. This is called on parent and `obtain_vsock_cid` returns cid, not port. So don't we always expect it to be `2`?

sardok: Ah, ok this enclave's main.

vsock-proxy/enclave/src/main.rs
```diff
@@ -16,15 +16,47 @@ use std::process;
...
+
+    let fd = nix::fcntl::open(constants::VSOCK_DEVICE_PATH, OFlag::O_RDONLY, Mode::empty())
+        .map_err(|e| format!("unable to open vsock device: {e}"))?;
+
+    let mut cid: u32 = 0;
+    unsafe { get_local_cid(fd, &mut cid).map_err(|e| format!("unable to get local cid: {e}"))? };
+
+    debug!("Obtained CID: {cid}");
+
+    Ok(cid as u32)
```
sardok: i don't think casting is needed here?

vsock-proxy/parent/src/platform/mod.rs
```diff
@@ -9,16 +9,15 @@ use tokio::task::JoinHandle;
...
 #[cfg(platform = "simulator")]
 mod simulator;
 #[cfg(platform = "snp")]
 mod snp;
 #[cfg(platform = "tdx")]
 mod tdx;
 
-#[cfg(any(platform = "snp", platform = "tdx", platform = "simulator"))]
-mod qemu;
-
```
sardok, line 21: i think these modules declerations are sorted by module name, any reason to move up?

vsock-proxy/parent/src/platform/qemu.rs
```diff
@@ -237,6 +276,65 @@ fn parse_gpu_bdfs(value: Option<&str>) -> Vec<String> {
...
+                return Ok((
+                    VmConnectionConfig {
+                        parent_vsock_listener: parent_socket,
+                        guest_vsock_cid: cid,
+                    },
+                    fd,
+                ));
+            }
+            Err(err_code) => {
+                let _ = nix::unistd::close(fd);
```
sardok, line 323: In this case what happens to `parent_socket`? Does it require explict closing?

sardok: Ok, `Drop` impl seems to be closing it: https://github.com/rust-vsock/vsock-rs/blob/56852d4006efce337875594e6a21ec49f0fc9c89/src/lib.rs#L214

gilanghamidy: Yeah I assume Rust library follows already the RAII principle, so it should close the underlying resources whenever it leaves the scope without moving the object.

vsock-proxy/parent/src/platform/mod.rs
gilanghamidy: The `sanity.sh` nags about it :sweat_smile: I remember someone told me (maybe @tvsfx) about the Edition 2024 `cargo fmt` that differs with previous editions that makes confusion between our toolings and scripts.

vsock-proxy/enclave/src/main.rs
gilanghamidy: Ah yeah true. Previously I think I wrote using `i32` instead of `u32`. I maybe changed that part but forgot to change this part.

vsock-proxy/enclave/src/main.rs
gilanghamidy: Ah I didn't know this macro before, I'll just change it to this as it is for the exact purpose that I am trying to do here

vsock-proxy/parent/src/platform/mod.rs
tvsfx: I doubt that this is true; I don't think the style edition had impact on module sorting, and rustfmt tends to leave blank lines alone.

If I create the following test file:
```rust
mod b;
mod c;

mod a;
```
then `rustfmt main.rs --style-edition="<any edition here>" --emit stdout` does nothing, and neither does `cargo fmt --check`

sardok: Interesting, `cargo fmt` doesn't seem to care.

sardok (approved):

vsock-proxy/parent/src/platform/mod.rs
gilanghamidy: I'll just follow whatever `sanity.ci` wants as it is Gandalf that needs to let me pass the PR

gilanghamidy:
I will merge now, additional refactor can follow further down the line.

PR #138, opened 2026-08-24
[RTE-1154] Filter Hosts file before writing to Enclave Host file

pavankalyan0424:
This PR implements filtering of /etc/hosts before writing to enclave.
Procress: Read all entries from /etc/hosts on parent, filter for any restricted hosts and then write the filtered hosts in enclave hosts file.

### DIFF
--- a/vsock-proxy/enclave/src/enclave.rs
+++ b/vsock-proxy/enclave/src/enclave.rs
@@ -42,8 +42,8 @@ use futures::{AsyncBufReadExt, StreamExt};
 use log::{debug, error, info, warn};
 use nix::net::if_::if_nametoindex;
 use shared::models::{
-    ApplicationConfiguration, GlobalNetworkSettings, NBDConfiguration, NetworkDeviceSettings,
-    PrivateNetworkDeviceSettings, SetupMessages, UserProgramExitStatus,
+    ApplicationConfiguration, GlobalNetworkSettings, HostEntries, NBDConfiguration,
+    NetworkDeviceSettings, PrivateNetworkDeviceSettings, SetupMessages, UserProgramExitStatus,
 };
 use shared::netlink::arp::NetlinkARP;
 use shared::netlink::route::NetlinkRoute;
@@ -82,7 +82,7 @@ const CERT_RENEWAL_INTERVAL_RELEASE: Duration =
     Duration::from_secs(24 * 60 * 60 /* 24 hours */);
 const CERT_RENEWAL_INTERVAL_DEBUG: Duration = Duration::from_secs(20 /* 20 sec */);
 
-const NETWORK_FILE_PATH_ALLOW_LIST: [&str; 3] = [DNS_RESOLV_FILE, HOSTNAME_FILE, HOSTS_FILE];
+const NETWORK_FILE_PATH_ALLOW_LIST: [&str; 2] = [DNS_RESOLV_FILE, HOSTNAME_FILE];
 
 fn default_cert_dir() -> PathBuf {
     PathBuf::from(ENCLAVE_FS_OVERLAY_ROOT)
@@ -839,6 +839,7 @@ async fn setup_enclave_networking(
         .map_err(|err| format!("Failed creating /run/resolvconf. {:?}", err))?;
 
     write_network_files(&global_settings, &NETWORK_FILE_PATH_ALLOW_LIST)?;
+    write_hosts_file(&global_settings.host_entries)?;
     debug!("Enclave global network settings files have been created.");
 
     enable_loopback_network_interface()?;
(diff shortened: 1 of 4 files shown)


pavankalyan0424:
Test Results

Docker run cmd:
```
docker run  --hostname abc  --add-host api:127.0.0.1 --add-host api:127.0.0.5 --add-host localhost:10.0.0.1 --add-host abc:10.0.0.1 ...
```

Previously:
```
===== /etc/hosts content =====
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::	ip6-localnet
ff00::	ip6-mcastprefix
ff02::1	ip6-allnodes
ff02::2	ip6-allrouters
127.0.0.1	api
127.0.0.5	api
10.0.0.1	localhost
10.0.0.1	abc
172.17.0.3	abc

===== end /etc/hosts =====
```

After my change:
```
===== /etc/hosts content =====
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::	ip6-localnet
ff00::	ip6-mcastprefix
ff02::1	ip6-allnodes
ff02::2	ip6-allrouters
10.0.0.1	abc
172.17.0.3	abc
127.0.0.1	api
127.0.0.5	api

===== end /etc/hosts =====
```

vsock-proxy/enclave/src/enclave.rs
```diff
@@ -849,6 +850,25 @@ async fn setup_enclave_networking(
...
+            ::1\tlocalhost ip6-localhost ip6-loopback\n\
+            fe00::\tip6-localnet\n\
+            ff00::\tip6-mcastprefix\n\
+            ff02::1\tip6-allnodes\n\
+            ff02::2\tip6-allrouters\n",
+    );
+    for (ip, hostnames) in host_entries {
+        output.push_str(&ip.to_string());
+        output.push('\t');
+        output.push_str(&hostnames.join(" "));
```
jethrogb: Need to validate contents of `hostnames` before writing to file

pavankalyan0424: At the parent end, I am currently handling this with
```
if hosts.iter().any(|host| RESTRICTED_HOSTS.contains(host)) {
            continue;
        }
```
Do you think we should consider any additional checks?

jethrogb: The enclave can't rely on anything done by the parent.

pavankalyan0424: Added a check on the enclave side, to validate whether hostname is valid.

vsock-proxy/enclave/src/enclave.rs
```diff
@@ -849,6 +850,60 @@ async fn setup_enclave_networking(
...
+    }
+
+    if !host_bytes[host_bytes.len() - 1].is_ascii_alphanumeric() {
+        return false;
+    }
+
+    host_bytes
+        .iter()
+        .all(|&b| b.is_ascii_alphanumeric() || b == b'-' || b == b'.')
+}
```
sardok, line 888: Also there is a note about its size as well;

From man-7 hostname:
```
       Each element of the hostname must be from 1 to 63 characters long
       and the entire hostname, including the dots, can be at most 253
       characters long.  Valid characters for hostnames are ASCII(7)
       letters from a to z, the digits from 0 to 9, and the hyphen (-).
       A hostname may not start nor end with a hyphen.
```
Also this limitation is mentioned in https://man7.org/linux/man-pages/man5/hostname.5.html too.

jethrogb: I think the DNS specification is a better source of truth.

pavankalyan0424: https://man7.org/linux/man-pages/man5/hostname.5.html is related to /etc/hostname, should we apply same rule for hostname entry in /etc/hosts?

vsock-proxy/enclave/src/enclave.rs
```diff
@@ -849,6 +850,60 @@ async fn setup_enclave_networking(
...
+        return false;
+    }
+
+    if !host_bytes[0].is_ascii_alphabetic() {
+        return false;
+    }
+
+    if !host_bytes[host_bytes.len() - 1].is_ascii_alphanumeric() {
+        return false;
+    }
```
jethrogb: Prefer accessor functions instead of manual index calculation and indexing. This also lets you do the is_empty check in one go.

```rust
match (host_bytes.first(), host_bytes.last()) {
    (Some(first), Some(last)) if first.is_ascii_alphabetic() && last.is_ascii_alphabetic() => {},
    _ => return false
}
```

vsock-proxy/enclave/src/enclave.rs
pavankalyan0424: I checked some RFC documents, based on them
1. https://datatracker.ietf.org/doc/html/rfc2181#page-13 => "That one restriction
   relates to the length of the label and the full name.  The length of
   any one label is limited to between 1 and 63 octets.  A full domain
   name is limited to 255 octets (including the separators)."
   
   
2. https://datatracker.ietf.org/doc/html/rfc1123#page-12
"The syntax of a legal Internet host name was specified in [RFC-952](https://datatracker.ietf.org/doc/html/rfc952)
      [DNS:4].  One aspect of host name syntax is hereby changed: the
      restriction on the first character is relaxed to allow either a
      letter or a digit.  Host software MUST support this more liberal
      syntax."

3. Other lexical grammar - https://datatracker.ietf.org/doc/html/rfc952

these are matching with what Sinan as pointed out.

vsock-proxy/enclave/src/enclave.rs
```diff
@@ -849,6 +851,72 @@ async fn setup_enclave_networking(
...
+
+    // labels are separated by dots ("."). https://datatracker.ietf.org/doc/html/rfc1034
+    let mut last_label = "";
+    for label in host.split('.') {
+        if label.len() > MAX_HOSTNAME_LABEL_LEN {
+            return false;
+        }
+
+        // alphabet (A-Z), digits (0-9), minus sign (-). The first and last character must be an alpha character
```
sardok: nit; `alpha character` -> `alpha-numeric character`.

sardok (dismissed):

sardok (approved):

PR #140, opened 2026-08-28
[RTE-1097] Skip mounting and unmounting the tmp dir when it does not exist

pavankalyan0424:
This PR implements the following change: skip mounting and unmounting the `tmp` directory when it does not exist.


aditijannu (dismissed):
LGTM. It would be good to add unit tests for obtaining `fetch_fs_mount_options` to ensure it works as expected. If possible, also create unit tests for `mount/unmount_file_system_nodes` or add an end to end test.

aditijannu (approved):

PR #144, opened 2026-09-04
[RTE-932] Harden Parent Base Image

gilanghamidy:
We are reducing the amount of packages installed in the parent-base to a minimum required. The package removed are related to debugging features that we may want to add separately in a debug converted images later on.


rui-alm (approved):

pavankalyan0424 (approved):

gilanghamidy:
Merging this. I retested with converted `nginx` and it is working as expected. (Able to serve HTTP page)

PR #145, opened 2026-09-04
[RTE-1053]: Pass cpu_count during snp measurements calculation.

sardok:
This PR includes `cpu_count` passed with conversion request during calculating measurements for SNP.


gilanghamidy (approved):
I missed this part when declining the RTE-1053 ticket. Thanks for quickly fixing it!

PR #147, opened 2026-09-08
[RTE-1148] Kernel cmdline hardening

pavankalyan0424:
This PR implements the following change - For prod/release builds Serial console is disabled and for debug builds serial console is enabled.


tools/container-converter/src/image_builder/enclave/qemu.rs
```diff
@@ -117,4 +117,13 @@ pub(crate) trait QemuEnclaveImageBuilder<'a> {
         self.compute_launch_measurements(&enclave_settings, &initramfs_file_path)
             .await
     }
+
+    #[allow(unused)]
+    fn kernel_cmdline(&self) -> &'a str {
+        if cfg!(debug_assertions) {
```
rui-alm: I'm considering the need for having the serial console on a "release" version of the converter.

Users will only have access to release versions of the converter, but have the ability to produce debug images, as in, setting the `debug` option on the conversion request.

If an issue happens during start up in production, we could ask a user to produce a debug image so we could see the missing logs.

@aditijannu Do you think we should rely on the conversion request `debug` to enable the serial console instead?

tools/container-converter/src/image_builder/enclave/qemu.rs
```diff
@@ -117,4 +117,13 @@ pub(crate) trait QemuEnclaveImageBuilder<'a> {
...
+            return "console=ttyS0 rdinit=/init loglevel=7";
+        } else {
+            return "console=null rdinit=/init loglevel=7";
```
sardok: Nit you can leave `return` and `';` out to return the last expression.

tools/container-converter/src/image_builder/enclave/qemu.rs
```diff
@@ -117,4 +117,13 @@ pub(crate) trait QemuEnclaveImageBuilder<'a> {
         self.compute_launch_measurements(&enclave_settings, &initramfs_file_path)
             .await
     }
+
+    #[allow(unused)]
+    fn kernel_cmdline(&self) -> &'a str {
```
sardok: Nit, this can be `&'static str`.

tools/container-converter/src/image_builder/enclave/qemu.rs
pavankalyan0424: This is how we do for nitro
```
pub(crate) fn is_enclaveos_debug_enabled() -> bool {
    env::var(ENCLAVEOS_DEBUG_ENV).as_deref() == Ok(ENCLAVEOS_DEBUG_VALUE)
}
...

if is_enclaveos_debug_enabled() {
        args.push("--debug-mode");
    }
```
 So I tried same. But open for thoughts

aditijannu: @rui-alm yes it's a good idea to enable serial console only via `debug` flag in the conversion request. This also changes the measurements.

@pavankalyan0424 For nitro, the nitro-cli run command was not really part of the measurement calculated in the converter. `--debug-mode` is more of a runtime option and I think it results in some PCRs being reset to 0 (not sure).

pavankalyan0424: @aditijannu I made changes at both places, so measurements are matching. But yeah let me try implementing debug field from conversion request.

aditijannu: I think it's still possible to leak some logs until you hit the measurement mismatch logic when using env var, or if you don't configure an app cert, it will not be detected at all. We should also update the behavior on the nitro side in the future.

pavankalyan0424: Done. I have changed my approach - to use the debug field from conversion request. Pls review it again. Thanks!

vsock-proxy/parent/src/platform/nitro.rs
sardok, line 30: Btw, does it seem wrong? `nitro` enclave forwards client logs when enclaveos_debug is not set.

tools/container-converter/src/image_builder/parent/qemu.rs
```diff
@@ -193,6 +196,10 @@ pub(crate) trait QemuParentImageBuilder<'a> {
         format!("MEM_SIZE={}M", mem_size)
     }
 
+    fn debug_env_var(&self) -> String {
+        format!("DEBUG={}", self.is_debug().unwrap_or_default())
```
sardok, line 200: This ok but we have `ENCLAVEOS_DEBUG_ENV` too. I am wondering if it makes better sense to leverage `ENCLAVEOS_DEBUG` here, instead of introducing a new env varible. Like,
`ENCLAVEOS_DEBUG=info` app logs (when conversion request debug is false)
`ENCLAVEOS_DEBUG=trace` app + kernel logs (when conversion request debug is true).

tools/container-converter/src/image_builder/enclave/qemu.rs
```diff
@@ -117,4 +117,13 @@ pub(crate) trait QemuEnclaveImageBuilder<'a> {
         self.compute_launch_measurements(&enclave_settings, &initramfs_file_path)
             .await
     }
+
+    #[allow(unused)]
```
sardok, line 121: Is this necessary?

tools/container-converter/src/image_builder/parent/qemu.rs
pavankalyan0424: I agree, but currently ENCLAVEOS_DEBUG is consumed by nitro. I don't want to disturb it. But let me check on this again.

tools/container-converter/src/image_builder/enclave/qemu.rs
pavankalyan0424: Simulator mode complains if it is not there.

vsock-proxy/parent/src/platform/nitro.rs
pavankalyan0424: Yeah agree. @aditijannu any idea why this was set? can I update this behaviour?

vsock-proxy/parent/src/platform/tdx.rs
```diff
@@ -92,28 +92,35 @@ mod tests {
...
-        diff_args(&expected, &Vec::from_iter(args.iter().map(String::as_str)));
+        for is_debug in [true, false] {
+            let expected_cmdline = if is_debug {
+                "console=ttyS0 rdinit=/init loglevel=7"
+            } else {
+                "console=null rdinit=/init loglevel=7"
+            };
+            // Ignore formatting to keep logical key/value pairs align better in a single line.
+            #[rustfmt::skip]
+            let expected = vec![
```
dlegalov-fortanix: Nit: It might be good to extract the creation of this vector to a function to reduce code duplication and prevent issues related to that.

tools/container-converter/src/image_builder/parent/qemu.rs
pavankalyan0424: I completely changed the approach. I am now passsing debug mode as argument instead.

vsock-proxy/parent/src/platform/snp.rs
dlegalov-fortanix, line 108: This vector seems to be, in large part, reused in tests and shared between platforms (barring tdx/snp specific parameters). Perhaps is might be beneficial to have a function return the parameters as needed for each platform, so that there is less code duplication? This is only  a suggestion.

vsock-proxy/parent/src/platform/nitro.rs
dlegalov-fortanix: Has this question been addressed?

dlegalov-fortanix (approved):

vsock-proxy/parent/src/platform/nitro.rs
pavankalyan0424: I checked the git history and I see the behaviour is correct. See https://github.com/fortanix/salmiac/commit/89883a6e0ba4c8f547954f7b693202a68ca4d73b#diff-53e369f321d0fc291039f9e58b17fa9f5e82a3bb11b3fafd311f9f0a3cdd2a0bL150. May be the name is confusing.

vsock-proxy/parent/src/platform/snp.rs
pavankalyan0424: Denis, I think this is intentional. We can update it later, if we are having more usecase around it.

rui-alm (approved):

PR #149, opened 2026-09-08
[RTE-1158] Introducing mechanics to structurize resolv.conf file

gilanghamidy:
- Doing all parsing of `resolv.conf` file in the parent, and pass that into the enclave in our own data structure, instead of plainly passing the `resolv.conf` file
- Removing the the mechanics of passing a full file directly to an enclave from the host as it is not being used elsewhere anymore


gilanghamidy:
Build is succesful in our internal CI, which I assume the `vsock-proxy` changes that also affects Nitro should also working correctly.

vsock-proxy/shared/src/models.rs
```diff
@@ -89,16 +91,135 @@ pub type HostEntries = HashMap<IpAddr, Vec<String>>;
...
-
     pub host_entries: HostEntries,
+    pub resolv_config: ResolvConfig,
+}
+
+/// Data structure that represents the content of /etc/resolv.conf file from
+/// the parents that we want to pass to the enclave. It is following the description
+/// from: https://man7.org/linux/man-pages/man5/resolv.conf.5.html
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ResolvConfig {
```
tvsfx, line 102: Did you check out https://crates.io/crates/resolv-conf ? It's co-maintained by Dirkjan Ochtman, a very reputable Rust dev, and it's used in https://github.com/hickory-dns/hickory-dns

gilanghamidy: Yeah I was wondering about that, but that crate has no `serde` integration and basically handling the parsing the `resolv.conf` file in a different way, (e.g., expanding all option fields into its own `bool`). Interfacing with its library for this purpose is way more painful than implementing a simple parser that performs simple sanitization, since we are not interested in actually consuming the config file.

sardok: +1 `resolv-conf` also does ip address validation: https://github.com/hickory-dns/resolv-conf/blob/3093f0b0bc37e5dff5dc13ab523a320bca9bd6ea/src/ip.rs#L15.

tvsfx: That crate does have a lot more validation in parsing overall. I also prefer the `bool` fields or a map/set here, since it avoids the question of what to do with duplicates by design. For `Display`, I also happened to see this implementation: https://docs.rs/domain/latest/src/domain/resolv/stub/conf.rs.html#484-590 . I think your approach of being verbose and always printing the config value is fine though.

vsock-proxy/enclave/src/enclave.rs
```diff
@@ -926,18 +925,62 @@ fn write_hostname_file(hostname: &str) -> Result<(), String> {
...
-    allow_list: &[&str],
-) -> Result<(), String> {
-    for file in global_settings
-        .global_settings_list
-        .iter()
-        .filter(|file| allow_list.contains(&file.path.as_str()))
-    {
-        write_to_file(Path::new(&file.path), &file.data, &file.path)?;
-        debug!("Successfully created {} inside an enclave.", &file.path);
+fn write_resolv_conf(resolv_conf: &ResolvConfig) -> Result<String, String> {
```
sardok, line 928: The parent is untrusted. So the validation should take place on enclave side.

sardok: Also if the use case only includes `nameserver IP`, then its ok to receive nameservers only as part of `GlobalNetworkSettings` and do validation on enclave side.
If we need more than nameserver ip, i am in favor of passing the content of `resolve.conf` as is, and do strong validation here via `resolv-conf` crate.

gilanghamidy: Yeah I understand. But I don't do excessive validation here because in the end, the file will be consumed by the `resolv` daemon anyway, and it will do entirely proper validation including whether the `nameserver` is a correct IP address or not. What I aim with this PR is to remove plaintext transfer between untrusted hosts and the enclave, and do minimal sanitization, especially filtering the comments or unacceptable line. But if you think someone can still put malicious shellcode within the `nameserver`, `search`, and `sortlist` arguments, then I'll change into full validation that is available in that crate.

gilanghamidy:
Maybe now I'll tag @rui-alm if it is preferrable to use the `resolv-conf` crate or implement ourselves here. To summarize:
1. `resolv-conf` crate does not have `serde` compatibility, requiring us to rewrite the entire structure and its `Into` and `From` implementation if we want to pass full structure in the `GlobalNetworkConfig` structure.
2. If we want to still use `resolv-conf` crate but in a simpler way, I propose to revert into the previous "accept entire file", and let the enclave runtime parse and validate that file first, before rewriting it into its local environment. In this way, it simplifies by not requiring to create our own verbose data structure, but still have the security guarantee with the parsing/validation, and excluding unnecessary comments that may be abused.

rui-alm:
The main goal is to remove the code that allows writing full files into the enclave.

The validation of resolv.conf ultimately needs to be done in the enclave. If there is a crate that does a good job at it, we can use it to read/write and validate the configuration. I haven't inspected how the crate provides the configuration values, but if possible, we can map them to our `serde` compatible `ResolvConfig` on the parent, send to the enclave and then use the crate to validate the config.

If there's no way to obtain the configuration through the crate, and the only way is to obtain the whole file, then we can keep the parent behavior with `ResolvConfig`, pass the structure into the enclave, write the resolv.conf file and then use the crate to validate the file in the enclave.

gilanghamidy:
@rui-alm to make a serde compatible one, we need to copy paste the field one by one. I think it is doable, just a bit verbose. I'll look into it ASAP after finishing early implementation of my other ticket

sardok:
I don't think serde is needed? We can use resolv-conf on both parent & enclave.

on parent in `customize_resolv_conf`:
* Parse from file i.e `Config::parse(parent_resolv)`.
* Iterate nameservers replace `ScopedIp`'s starting  `127.0.0.` with `nameserver_address`
* Then dump it to a string and send it via `ResolvConfigResult`.

Those steps would simplify `customize_resolv_conf`, now it is a bit messy IMO.

on enclave:
* Just `Config::parse` & and dump it to a file (`parse()` validates).

Don't understand why we need intermediate data structure where we can use `ResolvConfResult`.

vsock-proxy/shared/src/models.rs
```diff
@@ -89,16 +91,135 @@ pub type HostEntries = HashMap<IpAddr, Vec<String>>;
...
+    /// Lists of nameservers, in the order of appearance
+    pub nameservers: Vec<String>,
```
rui-alm, line 104: We can be more restrictive here:
> `nameserver` Name server IP address
              Internet address of a name server that the resolver should
              query, either an IPv4 address (in dot notation), or an IPv6
              address in colon (and possibly dot) notation as per RFC
              2373.  Up to MAXNS (currently 3, see <res_state.h>) name
              servers may be listed, one per keyword.

The `nameserver` is an IP address so we can represent it as `IpAddr` instead of `String`.

rui-alm (approved):
I'd prefer to have a crate doing the parsing and validation, but we can assess that separately. This PR remove the `FileWithBytes` writing functions and defines a rigid structure of what can be passed into the enclave. It restricts the arbitrary data that can be written to the file, although we should restrict nameservers to IP addresses.

I'm fine with these improvements being done in a next stage.

vsock-proxy/shared/src/models.rs
gilanghamidy: I'll mark that as an improvement. There are some logics need to be changed w.r.t. checking if the specified address is loopback. I was following the original naive check using `starts_with` function on a string, which definitely needs to be changed as well.

gilanghamidy:
Merging with follow up ticket to implement more defense in depth

PR #151, opened 2026-09-09
[RTE-978]: Replace nvidia ubuntu packages with nvidia-artifacts.

sardok:
This PR ends using ubuntu repositories for nvidia packages and uses nvidia-artifacts file instead.

Now `nvidia-artifacts.tar.gz` is expected in `docker/common-builder-context` directory.

### DIFF
--- a/.github/workflows/build-docker-images.yml
+++ b/.github/workflows/build-docker-images.yml
@@ -1,4 +1,4 @@
-name: Build and publish Docker Images
+name: Build Docker Images
 
 on:
   push:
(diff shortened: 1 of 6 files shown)


.github/workflows/build-docker-images.yml
```diff
@@ -27,7 +27,8 @@ jobs:
       - name: Build docker images
         run: |
           docker build --no-cache docker/enclave-base -t enclave-base
-          docker build --no-cache docker/enclave-base-gpu -t enclave-base-gpu
+          # Disabled due to required artifacts
+          # docker build --no-cache docker/enclave-base-gpu -t enclave-base-gpu
```
pavankalyan0424, line 31: @sardok Is there a possibility of using dummy nvidia tar file for CI? may be we could create something as part of steps

sardok: i don't think so, `stage-nvidia-driver-payload.sh` fails if the content of tar.gz file is unexpected.

sardok: There will be another new artifact (`busybox`) that will be used by enclave-base & enclave-base-gpu (probably parent too). We need to find a more structured way to handle those artifacts in CI but we need to make those artifacts to be publicly available first.

pavankalyan0424 (dismissed):

.github/workflows/build-docker-images.yml
pavankalyan0424: ok, got it. Yeah we can take it in another follow up PR.

rui-alm (dismissed):

pavankalyan0424 (approved):

PR #154, opened 2026-09-10
[RTE-1176] Document unsupported converter options

aditijannu:
- Cleanup and update converter's openapi documentation
- Include error messages and unit tests for unsupported configurations available.

### DIFF
--- a/api-model/docs/models.openapi.yaml
+++ b/api-model/docs/models.openapi.yaml
@@ -284,12 +284,22 @@ components:
     SNPEnclavesConversionResponse:
       type: "object"
       required:
-        - converted_image
+        - name
+        - sha
+        - size
         - config
       properties:
-        converted_image:
-          $ref: "#/components/schemas/ConvertedImageInfo"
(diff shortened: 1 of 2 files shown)


api-model/docs/models.openapi.yaml
```diff
@@ -320,22 +284,12 @@ components:
...
-          type: "string"
-          description: "Converted image sha"
-          format: hex
-        size:
-          type: "integer"
-          format: "int64"
-          description: "Converted image size"
+        converted_image:
+          $ref: "#/components/schemas/ConvertedImageInfo"
+          description: "Reference properties of the converted image"
```
sardok: `converted_image` is flattened in the model.
see it here: https://github.com/fortanix/salmiac/blob/061760e8681de3af7a2da7a8780cec978c7e9458/api-model/src/snp.rs#L50-L57

So i believe `name`, `sha` & `size` are correct fields here.

sardok (approved):

PR #155, opened 2026-09-11
[RTE-930]: Revisit vsock-proxy protocol messages.

sardok:
This PR revisit the strings/errors exchanged between parent and enclave. The motivation is to prevent leaking sensitive information from enclave.

Summary of changes:
* Introduced `CertificateErrorCode` for finer error propagation to enclave
* `UserProgramExit` does not contain `String` as error now. Instead it receives a new type `EnclaveErrorCode`.
* Added `EnclaveErrorCode` type. There is not much in it but we can add more invariants in it to describe what went wrong or at what stage.

We loose some information here, for example now `SetupMessages::invariant_name()` is used as opposed to full `SetupMessages` invariant dump, or `UserProgramExit` error string (although we will probably observe what went wrong in enclave logs), so this PR may make debugging harder. Letme know if you think, some of the changes in this PR should be reverted.


vsock-proxy/shared/src/lib.rs
```diff
@@ -175,10 +175,10 @@ macro_rules! extract_enum_value {
     ($value:expr, $pattern:pat => $extracted_value:expr) => {
         match $value {
             $pattern => Ok($extracted_value),
-            x => Err(format!(
-                "Expected {:?} for enum variant, but got {:?}",
+            e => Err(format!(
+                "Expected {} for enum variant, but got {}",
                 stringify!($pattern),
-                x
+                e.variant_name(),
```
sardok, line 181: `extract_enum_value` is used in many places on enclave. Some of them were probably causing `UserProgramExit` where the string here was propagated.
Since we don't propagate full error strings to via `UserProgramExit`, i am thinking if `extract_enum_value` should dump full invariant here (as it used to) for easier debugging.

vsock-proxy/shared/src/models.rs
```diff
@@ -22,7 +22,7 @@ pub enum SetupMessages {
     GlobalNetworkSettings(GlobalNetworkSettings),
     CSR(String),
     Certificate(String),
-    UserProgramExit(Result<UserProgramExitStatus, String>),
+    UserProgramExit(Result<UserProgramExitStatus, EnclaveErrorCode>),
```
sardok, line 25: This is the main change of this PR.

vsock-proxy/shared/src/lib.rs
rui-alm: We can consider more verbose errors for debug builds, but restricting (and when possible defining) what gets logged for production makes sense.

For the parent code we could log more if necessary, but given there is common code we may have to be conservative on return values and introduce explicit logging in common parts that would get suppressed for the enclave, for example, debug logging that could be enabled for the parent but restricted for the enclave except for debug builds.

vsock-proxy/shared/src/netlink/route.rs
```diff
@@ -194,7 +193,9 @@ impl TryFrom<&RouteMessage> for Route {
         fn v4_network(address: Option<(IpAddr, u8)>) -> Result<Option<Ipv4Network>, String> {
             match address {
                 Some((addr, prefix)) => {
-                    let ipv4 = extract_enum_value!(addr, IpAddr::V4(e) => e)?;
+                    let IpAddr::V4(ipv4) = addr else {
+                        return Err("Unexpected ipv6 address".to_owned());
```
rui-alm: Should we be more clear and state that an IPv4 address was expected but got an IPv6 instead? I'm afraid someone will interpret this as if an invalid (but expected) IPv6 address was passed.

vsock-proxy/parent/lib/lib.rs
```diff
@@ -58,34 +58,65 @@ pub async fn handle_csr_message<
...
-            // waiting.
-            vsock.write_lv_bytes(&[]).await?;
-            Err(e)
+        Err(code) => {
+            // Always reply, including when NODE_AGENT is absent, so the enclave cannot
+            // remain blocked waiting for a certificate response and can retry after some time.
+            vsock
+                .write_lv(&SetupMessages::CertificateError(code))
+                .await?;
+            info!("Error requesting App Certificate: {:?}", code);
```
rui-alm: Should this be an error log?

vsock-proxy/parent/lib/lib.rs
```diff
@@ -58,34 +58,65 @@ pub async fn handle_csr_message<
...
+    vsock: &mut Socket,
+    cert_api: CertApi,
+    csr: String,
+    address: Option<String>,
+    timeout: Duration,
+) -> Result<(), String> {
+    let result = match address {
+        None => Err(CertificateErrorCode::Unavailable),
+        Some(address) => {
+            info!("Requesting CCM for App Certificate, timing out after 60 sec...");
```
rui-alm: Should we replace the hardcoded 60 seconds with the value of `timeout`?

vsock-proxy/shared/src/lib.rs
sardok: Makes sense. Fixed the code. Now there are two variants of `extract_enum_value`. Debug version logs full error type.

vsock-proxy/shared/src/netlink/route.rs
sardok: You're right. Fixed the error message. Hope its more clear now.

vsock-proxy/parent/lib/lib.rs
sardok: Ok, its `error!` now. Since this retryable error i am wondering if `warn` would fit here better ?

vsock-proxy/parent/lib/lib.rs
sardok: Yes of course. Fixed the message.

vsock-proxy/parent/lib/lib.rs
rui-alm: I guess it could be `warn` if we logged an `error` after exhausting the retries, but is there an actual retry mechanism that I'm missing or are you talking about the certificate renewal interval?

```/// The interval between certs are checked for renewal
const CERT_RENEWAL_INTERVAL_RELEASE: Duration =
    Duration::from_secs(24 * 60 * 60 /* 24 hours */);
const CERT_RENEWAL_INTERVAL_DEBUG: Duration = Duration::from_secs(20 /* 20 sec */); 
```

I don't see a maximum number of retries anywhere, but if we have one, we could switch these to warn and then log as error once we exhaust the retries.

If we retry indefinitely, then logging as warn may go unnoticed if the error just continues to happen, although it could be noticeable if there's not too many warnings and the time between retries is short enough.

sardok: Yes i've just revisited the code. Yes, you're right there is no retry mechanism. Everything relies on automatic renewal: `auto_cert_renewals`. So this error message will be printed on parent side, enclave part `auto_cert_renewal` should print its own error message.

sardok: Btw, after being printed as `error` this failure will turn into `info` message again. See here:

https://github.com/fortanix/salmiac/blob/8e1beb9cbc66aff93e2a9d81b21840f2333574ec/vsock-proxy/parent/src/parent.rs#L83

rui-alm: It looks to me like this should be logged as an error, since the error is coming from `handle_csr_message`, not the enclave.

There were a few places where the assumption was that any error related with the CSR would be "fine" since the enclave would retry, but I think you fixed it now.

An error from the parent when reaching node agent, or an error coming from node agent, should be logged as errors, or warn+error when there's a max retries.

rui-alm:
I'm approving the PR, but feel free to switch that info log to error, but it's not a blocker, as the objective is to suppress sensitive data from the enclave. We should do a separate review to the whole logging (besides the enclave), their usefulness, log levels and areas where we should log more.

rui-alm (dismissed):

vsock-proxy/parent/lib/lib.rs
sardok: Agreed. Also changed `parent/src/parent.rs:83` info log to error log too.

rui-alm (approved):

PR #158, opened 2026-09-16
[RTE-1202]: Use em-client & sdkms client.

sardok:
This PR replaces em_app's usage for certain functionality with explicit em-client & sdkms client. With em-client & sdkms-client, we have luxury of passing underlying http client. The http-client in this PR uses rustls deliberately.


vsock-proxy/shared/src/models.rs
```diff
@@ -256,9 +256,9 @@ pub enum UserProgramExitStatus {
 
 /// Public failure codes crossing the enclave boundary.
 /// Should not contain internal error details.
-#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 pub enum EnclaveErrorCode {
-    EnclaveFailure,
+    EnclaveFailure(String),
```
sardok: I'll revert this change after i confirm the PR works as intended.

vsock-proxy/enclave/src/file_system.rs
```diff
@@ -134,7 +134,7 @@ pub(crate) async fn mount_read_write_file_system(
             conn_info,
             OVERLAY_FS_SECURITY_OBJECT_PREFIX.to_string(),
             DERIVATION_DATA_IV.to_string(),
-            Some(Arc::new(read_root_certificates())),
+            Some(Arc::new(read_root_certificates_mbedtls())),
```
pavankalyan0424, line 137: Should we update tools/enclaveos-encrypted-fs code too? Agree that we are about to disable File system persistence for SNP, TDX. So we might need to fix this code before enabling the flag.

pavankalyan0424 (dismissed):

vsock-proxy/enclave/src/utils/net.rs
```diff
@@ -0,0 +1,292 @@
...
+    use test_case::test_case;
+
+    use super::get_hyper_tls_connector;
+    use crate::app_configuration::EmAppCredentials;
+    use crate::certificate::create_signer_key;
+
+    const TIMEOUT: Duration = Duration::from_secs(10);
+
+    #[test_case("ccm.fortanix.com", 443)]
+    #[test_case("192.0.2.1", 8443)]
```
aditijannu, line 116: Why this IP address?

vsock-proxy/enclave/src/enclave.rs
```diff
@@ -264,6 +265,9 @@ pub(crate) async fn run(
         )
         .await?;
 
+        // Consume the parent's pending setup message before fallible app configuration.
+        let log_conn_addrs = extract_enum_value!(parent_guard.deref_mut().read_lv().await?, SetupMessages::AppLogPort(addr) => addr)?;
```
aditijannu, line 269: If you change the order of receiving SetupMessages on the enclave side, you will need to update the order of sending the data on the parent side.

vsock-proxy/enclave/src/file_system.rs
aditijannu: Is there something wrong with this code since you mentioned we need to fix it?

vsock-proxy/enclave/src/utils/net.rs
sardok: Its just a made up ip. Is there an preference of yours? something like `192.168.2.1`?

vsock-proxy/enclave/src/file_system.rs
sardok: I don't remember mentioning about a fix. Only thing i can say about it is, we have rustls now, and we can drop mbedtls for reduced dependency. But first we need to get rid of em-app completely.

vsock-proxy/enclave/src/enclave.rs
sardok: I think the order between parent-enclave is consistent (if not missing anything). This is moved above `setup_app_configuration` from below of it, to receive logs happen in `setup_app_configuration` (or at least get buffered). Letme know if my intention has no related to this change. Without re-ordering i think i was having `extract_enum_value!` error if something goes wrong in `setup_app_configuration`.

vsock-proxy/enclave/src/utils/net.rs
sardok: Btw, this test is not very useful. I added that after i had `EmClient::try_new_with_connector` failure due to the wrong `base_path`. I can remove this test.

vsock-proxy/enclave/src/app_configuration.rs
```diff
@@ -449,13 +453,14 @@ iy6KC991zzvaWY/Ys+q/84Afqa+0qJKQnPuy/7F5GkVdQA/lfbhi
...
-        let cert_list = MbedtlsList::<Certificate>::new();
+        let mut reader = Cursor::new(key.as_bytes());
+        let key = rustls::internal::pemfile::rsa_private_keys(&mut reader)
+            .expect("Failed parsing test private key")
+            .remove(0);
 
         EmAppCredentials {
-            certificate: Arc::new(cert_list),
-            key: Arc::new(pk),
+            certificate: Vec::new(),
```
aditijannu, line 462: I'm confused why we set certificate and root_certificate to an empty vec or none, won't it result in an error?

vsock-proxy/enclave/src/enclave.rs
aditijannu: since you have tested your branch I think you got the order right.

aditijannu (approved):

vsock-proxy/enclave/src/app_configuration.rs
sardok: `EmAppCredentials::new()` checks if certificate is empty or not. `mock()` doesn't use `EmAppCredentials::new()` because of that reason i assumed.

PR #159, opened 2026-09-17
[RTE-931] Use persistent storage only when requested during conversion time

pavankalyan0424:
Previously, peristent storage was created even if enable_overlay_filesystem_persistence field is set as false during the conversion time. This PR updates this behaviour - Skips creating the persistent storage, when flag is set as false.


pavankalyan0424:
Test result
NBD exports get filtered out now
```
[2026-09-17T15:54:56Z INFO  enclave::enclave] Device 2 i[2026-09-17T15:54:56Z DEBUG parent::parent] NBD config is 
            [generic]
                includedir = /etc/nbd-server/conf.d
                allowlist = true
                listenaddr = 10.0.0.1
        
                [enclave-fs]
                    authfile =
                    exportname = /opt/fortanix/enclave-os/Blockfile.ext4
                    readonly = true
                    port = 7777
[2026-09-17T15:54:56Z INFO  parent::parent] Spawned nbd server on port 7777 serving block file /opt/fortanix/enclave-os/Blockfile.ext4
s connected and ready.
```

Also, Blockfile-rw.ext4 and overlayfs directory in parent are not created
```
root@90e6d2c678ce:/# ls /opt/fortanix/enclave-os/
Blockfile.ext4  bzImage  initramfs.gz  nbd.config  original-parent.env  parent  start-parent.sh
```

tools/container-converter/src/lib.rs
```diff
@@ -370,6 +374,19 @@ fn validate_request(request: &ConversionRequest) -> Result<()> {
         });
     }
 
+    if let Some(enable_ovelay_fsp) = request
```
aditijannu, line 378: This disables filesystem persistence support on nitro platform too. We don't want to do that.

tools/container-converter/src/image_builder/parent/qemu.rs
```diff
@@ -181,6 +184,13 @@ pub(crate) trait QemuParentImageBuilder<'a> {
         )
     }
 
+    fn enable_overlay_fsp_env_var(&self) -> String {
+        format!(
+            "ENABLE_OVERLAY_FILESYSTEM_PERSISTENCE={}",
```
aditijannu: I don't understand why we do this? Why an env var?

vsock-proxy/enclave/src/file_system.rs
aditijannu: We can rename this function to indicate it only sets up the dmcrypt/encrypted mount?

vsock-proxy/parent/lib/lib.rs
```diff
@@ -36,6 +21,26 @@ pub struct NBDExportConfig {
...
+pub fn get_filtered_nbd_exports(overlay_fsp_enabled: bool) -> Vec<NBDExportConfig> {
+    let mut nbd_exports: Vec<NBDExportConfig> = Vec::new();
+    nbd_exports.push(NBDExportConfig {
+        name: "enclave-fs",
+        block_file_path: "/opt/fortanix/enclave-os/Blockfile.ext4",
+        port: 7777,
+        is_read_only: true,
+    });
+
+    if overlay_fsp_enabled {
```
aditijannu: Nit: Technically we always use overlay because we have dmverity as the ready only layer. This isn't your code changes but persistence_enabled may be a better variable name here and in other locations where we add conditional checks based on this boolean value.

vsock-proxy/parent/src/main.rs
```diff
@@ -46,18 +47,23 @@ struct ParentConsoleArguments {
...
     pub enclave_extra_args: Vec<String>,
+
+    pub enable_filesystem_persistence: bool,
 }
 
 impl ParentConsoleArguments {
     // 256MB converted to bytes
     const RW_BLOCK_FILE_DEFAULT_SIZE: u64 = 256 * 1024 * 1024;
 
+    const ENABLE_OVERLAY_FILESYSTEM_PERSISTENCE_ENV_VAR: &str =
```
aditijannu: Can we pass this as a console argument instead of using env var?

vsock-proxy/parent/src/parent.rs
```diff
@@ -96,12 +97,17 @@ async fn message_handler(enclave: &mut AsyncVsockStream) -> Result<UserProgramEx
...
-    info!("Checking presence of overlayfs parent directory.");
-    let overlayfs_parent_dir = Path::new(OVERLAYFS_BLOCKFILE_DIR);
-    if !overlayfs_parent_dir.exists() {
-        info!("Creating overlayfs directory where the rw encrypted blockfile would be created...");
-        fs::create_dir_all(overlayfs_parent_dir)
-            .map_err(|e| format!("Unable to create overlayfs parent dir : {:?}", e))?;
+    let overlay_fsp_enabled = args.enable_filesystem_persistence;
+    if overlay_fsp_enabled {
+        info!("Checking presence of overlayfs parent directory.");
+        let overlayfs_parent_dir = Path::new(OVERLAYFS_BLOCKFILE_DIR);
```
aditijannu: This is not your change but since you are updating the code here, its more appropriate to rename `OVERLAYFS_BLOCKFILE_DIR` to `ENCRYPTED_BLOCKFILE_DIR`, and `overlayfs_parent_dir` to `encryptedfs_parent_dir`

vsock-proxy/parent/lib/lib.rs
aditijannu: During your testing you can check how much space is available for the guest application to use - it should be in the order of mem_size passed at conversion time.

We could also update this info message when persistence_enabled to say `Encrypted space...` otherwise `Ephemeral space...`

vsock-proxy/parent/src/main.rs
aditijannu: Good to indicate with a warning message that this RW_STORAGE_SIZE will not be used if persistence is disabled

tools/container-converter/src/image_builder/parent/qemu.rs
pavankalyan0424: I needed a way to pass `enable_overlay_filesystem_persistence` to parent (here -https://github.com/fortanix/salmiac/pull/159/changes#diff-53e369f321d0fc291039f9e58b17fa9f5e82a3bb11b3fafd311f9f0a3cdd2a0bL98). (Enclave gets it from EnclaveSetting); But parent does not have a good way. I took inspiration from how ENABLE_GPU_PASSTHROUGH is passed to parent and how it is getting used ( See - https://github.com/fortanix/salmiac/blob/c7e7121580315b54057d827485b8aa7ead0e2322/vsock-proxy/parent/src/platform/qemu.rs#L182 )

vsock-proxy/parent/src/main.rs
pavankalyan0424: Yeah, I think I will need to update this logic. If I am passing a runtime env var with same name it is overwriting.

tools/container-converter/src/image_builder/parent/qemu.rs
pavankalyan0424: Based on comment here (https://github.com/fortanix/salmiac/pull/159#discussion_r4041376229), I am updating the behaviour.

aditijannu (approved):

